PR #7, opened 2020-09-19
interpolate player positions and rotations

piodul:
Now the game seems to stutter less, at least during good network conditions.


capturcus (approved):
lgtm

PR #8, opened 2020-09-19
wilhelm scream on death

kbr-:


capturcus (changes requested):
please change sound to play only when i killed

capturcus (approved):

PR #15, opened 2020-09-26
hidingspots name refactor (levelpacker.py)

mparafin:
Poprawka kontynuacyjna do refactoru nazw bush->hidingspot.


capturcus (approved):
did not work before this pr, works after

PR #16, opened 2020-09-29
dfctl-completion

mparafin:
Works after reloading bash. Updated README.md


capturcus (approved):
confirmed that this works with bash

PR #20, opened 2020-10-08
Level rework

mparafin:
A thorough refactoring of the level creation engine.


client/dfclient/src/gameplay_state.cpp
```diff
@@ -268,6 +307,25 @@ void GameplayState::OnMessage(const std::string& data) {
 
 	// draw remaining time
 	updateRemainingText(worldState->stepsRemaining());
+
+	// make current hidingspot transparent
+	if(worldState->hidingspot()->str() != "") {
```
capturcus: Won't this create a new tween every frame that the player is within the bounds of the hiding spot?

deadfish.fbs
```diff
@@ -7,16 +7,39 @@ struct Vec2 {
...
+table Tileinfo {
+  gid: uint16;
+  name: string;
+}
+
+table Object {
+    pos: Vec2;
     rotation: float;
+    gid: uint16;
+    hspot: bool;
```
capturcus: Is this bool necessary, or can we infer the hiding spot property from hspotname being not empty?

deadfish.fbs
```diff
@@ -108,6 +134,7 @@ table WorldState {
   mobs:[Mob];
   indicators:[Indicator];
   stepsRemaining:uint64;
+  hidingspot:string;
```
capturcus: Could we maybe change the name of this to currentHidingSpot or something like that?

levelpacker/levelpacker.py
```diff
@@ -1,71 +1,164 @@
...
+        polyverts = ""
+        poly = 0
 
+        if o.getElementsByTagName('ellipse'):
+            width = float(o.getAttribute('width'))
+            height = float(o.getAttribute('height'))
+            if width != height:
+                print("collision object of id", o.getAttribute('id'), "is an ellipse, not a circle, with width:", width, "height:", height)
+                exit(1)
+            ellipse = True
```
capturcus: Why is this set to true here? What does this variable do?

levelpacker/levelpacker.py
```diff
@@ -1,71 +1,164 @@
...
+        poly = 0
 
-def handle_meta(g: minidom.Node, objs: GameObjects, builder: flatbuffers.Builder):
+        if o.getElementsByTagName('ellipse'):
+            width = float(o.getAttribute('width'))
+            height = float(o.getAttribute('height'))
+            if width != height:
+                print("hidingspot", o.getAttribute('name'), "is an ellipse, not a circle, with width:", width, "height:", height)
+                exit(1)
+            ellipse = True
```
capturcus: likewise

levels/tilesets/default_tileset.tsx
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<tileset version="1.4" tiledversion="1.4.3" name="default_tileset" tilewidth="1200" tileheight="5760" tilecount="4" columns="0" objectalignment="topleft">
```
capturcus, line 2: Why is this file in the repository? After adding more art assets it will have to be changed every time

server/deadfish.cpp
```diff
@@ -0,0 +1,76 @@
...
+HidingSpot::HidingSpot(const FlatBuffGenerated::HidingSpot* fb_Hs) {
+    this->name = fb_Hs->name()->str();
+    b2BodyDef myBodyDef;
+	myBodyDef.type = b2_staticBody;
+	myBodyDef.position.Set(fb_Hs->pos()->x(), fb_Hs->pos()->y());
+	body = gameState.b2world->CreateBody(&myBodyDef);
+
+	b2FixtureDef fixtureDef;
+    fixtureDef.isSensor = true; // makes hiding spot detect collision but allow movement
+    if (fb_Hs->ellipse()) {
```
capturcus: If the ellipse variable makes the distinction between the fixture being a circle and a polygon maybe change it to "circle"? Or add an enum with the type of the shape

server/deadfish.cpp
```diff
@@ -0,0 +1,76 @@
...
+	body->SetUserData(this);
+}
+
+void HidingSpot::handleCollision(Collideable& other) {
+    try {
+        auto &player = dynamic_cast<Player &>(other);
+        playersInside.insert(&player);
+    }
+    catch (...) {
+        std::cout << "non-player collision with bush detected!" << std::endl;
```
capturcus, line 42: If I understand correctly, this is an unnecessary debug, if so, please remove it. The server output is too cluttered as it is

server/deadfish.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+Collision::Collision(const FlatBuffGenerated::Collision* fb_Col) {
+    b2BodyDef myBodyDef;
+	myBodyDef.type = b2_staticBody;
+	myBodyDef.position.Set(fb_Col->pos()->x(), fb_Col->pos()->y());
+    myBodyDef.angle = fb_Col->rotation() * TO_RADIANS;
+	body = gameState.b2world->CreateBody(&myBodyDef);
+
+	b2FixtureDef fixtureDef;
+    if (fb_Col->ellipse()) {
```
capturcus: likewise as with the previous "ellipse" flatbuffer variable

server/deadfish.hpp
```diff
@@ -100,19 +100,44 @@ struct Civilian : public Mob {
...
 };
 
+// just a data container to be able to send it later to clients
+struct Tileinfo {
+	Tileinfo(const FlatBuffGenerated::Tileinfo* fb_Ti) : name(fb_Ti->name()->str()), gid(fb_Ti->gid()) {}
+	std::string name;
+	uint16_t gid;
+};
+
+// just a data container to be able to send it later to clients
```
capturcus, line 110: Just one comment on the subject is probably informative enough :P sorry for being very nitpicky but dat shiet b messin w/ my ocd ni

server/level_loader.cpp
```diff
@@ -23,69 +23,67 @@ void initPlayerwall(const FlatBuffGenerated::PlayerWall *pw)
...
+	for (auto &o : gameState.level->objects) {
+		FlatBuffGenerated::Vec2 pos(o->pos.x, o->pos.y);
+		auto hspotname = builder.CreateString(o->hspotname);
+		auto offset = FlatBuffGenerated::CreateObject(builder, &pos, o->rotation, o->gid, o->hspot, hspotname);
+		objectOffsets.push_back(offset);
+	}
+	auto objects = builder.CreateVector(objectOffsets);
+
+	// decoration
+	std::vector<flatbuffers::Offset<FlatBuffGenerated::Decoration>> decorationOffsets;
```
capturcus, line 48: ugly and repetitive code that we can do nothing about without using macros because flatbuffers are mentally challenged

server/level_loader.cpp
```diff
@@ -23,69 +23,67 @@ void initPlayerwall(const FlatBuffGenerated::PlayerWall *pw)
...
+	for (auto &hs : gameState.level->hidingspots)
 	{
-		FlatBuffGenerated::Vec2 pos(b->body->GetPosition().x, b->body->GetPosition().y);
-		auto f = b->body->GetFixtureList();
-		auto c = (b2CircleShape *)f->GetShape();
-		auto off = FlatBuffGenerated::CreateHidingSpot(builder, c->m_radius, b->body->GetAngle(), &pos);
+		FlatBuffGenerated::Vec2 pos(hs->body->GetPosition().x, hs->body->GetPosition().y);
+		bool ellipse = false;
+		float radius = 0;
+		flatbuffers::Offset<flatbuffers::Vector<const FlatBuffGenerated::Vec2 *>> polyverts = 0;
```
capturcus: Why are you storing Vec2s here as pointers? The structures have only 2 floats in them, it adds complexity on account of having to use unique pointers

.gitignore
```diff
@@ -5,4 +5,5 @@ assets/
 client/data
 client/imgui.ini
 levelpacker/FlatBuffGenerated/
-levels/*.bin
\ No newline at end of file
+levels/*.bin
+*.tsbin
```
capturcus, line 9: What are .tsbin files used for? From what I've seen tilesets are currently embedded in the level data and there is no need for separate files (as it should be)

deadfish.fbs
```diff
@@ -7,16 +7,39 @@ struct Vec2 {
...
+}
+
+table Decoration {
     pos: Vec2;
+    rotation: float;
+    gid: uint16;
 }
 
-table HidingSpot {
+table Collision {
```
capturcus: The name is slightly misleading, as "Collision" doesn't mean the event of collision but rather "something that can be collided with". What do you think about renaming this to "Obstacle" or something like that?

capturcus (changes requested):
Apart from the review comments please fix the test map and make sure that the tests are executed properly. There is nothing wrong with removing old maps created for testing purposes, I suggest that you leave only the maps that are working with this rework in the repository.

client/dfclient/src/gameplay_state.cpp
mparafin: This is what the next `if` is for, two lines later :P Tweens will be created only if the sprites are at max opacity and they change the opacity immediately. The same goes for raising the opacity back up.

mparafin: Confirmed through dupadebugging.

levels/tilesets/default_tileset.tsx
mparafin: So it can be used immediately by anyone wanting to create a level. Maybe it should go in `data`, but I think it sholud stay in the repo.

server/deadfish.cpp
mparafin: This only happens if something really goes wrong, it doesn't show up if the server is working properly. Figured it could stay in case something breaks. Can be removed, but left intentionally

server/deadfish.hpp
mparafin: It shows up as the basic object documentation, like when you hover over the type name in IDE. In `deadfish.hpp` it may be OCD-inducing, but it's intentional as well :P

server/deadfish.cpp
capturcus: Won't this show up every time a mob enters a hiding spot?

mparafin: Nope, mob-hspot collisions are filtered by box2d. It shouldn't show up at all; if it does, that means that somebody configured something wrong regarding box2d, collision masks or something like that.

deadfish.fbs
mparafin: Done. Renamed to CollisionMask everywhere

.gitignore
mparafin: True, that's an uncleaned refactor artifact. Sry
```suggestion
```

server/level_loader.cpp
mparafin: That's what FlatBuffGenerated::CreateHidingSpot needed. But you have a point in that client doesn't need the hidingspots at all. Removed the section.

deadfish.fbs
mparafin: Yes we could. Done

deadfish.fbs
mparafin: We can indeed. Refactored.

server/level_loader.cpp
mparafin: We could try pushing it into a function with an additional parameter and iffing, but I think it would be much less readable and much more sore for the eyes. The copy-pasted operation isn't that long and that similar for this to make sense, imo.

server/deadfish.cpp
mparafin: It proved to be redundant, and thus was removed. `ellipse` is no more.

capturcus:
While deleting old maps you deleted the test map by mistake. Please make sure that the tests pass every time you're pushing to the repository by using the `make test` or `dfctl test` command (this is the first thing I do while considering a PR for merging). Please also modify the test map so that it includes the newly implemented level features so that they can be tested as well.

capturcus (changes requested):
.

mparafin:
Added testmap back, modified to include new features. Tests pass.

PR #21, opened 2020-10-10
tiles

mparafin:
What more to say, added tile mechanics and first basic tile sprites. `big.tmx` is a place to test out the background colors and pick one.
Appendix A: a color reference photo.
seabed


client/dfclient/src/gameplay_state.cpp
```diff
@@ -92,6 +95,40 @@ void GameplayState::LoadLevel() {
...
+				if (spritename.empty()) continue;
+				auto tileSprite = std::make_unique<ncine::Sprite>(this->cameraNode.get(), _resources.textures[spritename].get(),
+					j * TILE_WIDTH, -i * TILE_HEIGHT);
+				tileSprite->setAnchorPoint(0, 1);
+				tileSprite->setLayer(TILE_LAYER);
+				this->nodes.push_back(std::move(tileSprite));
+			}
+		}
+	}
+
```
piodul, line 115: Just my two cents: I feel we should avoid any custom parsing inside C++. Instead, we should parse the tilemap inside python and serialize it as a FlatBuffer-encoded array.

There is a dedicated library for parsing CSV included in python's stdlib which should make it easy: https://docs.python.org/3/library/csv.html

levelpacker/levelpacker.py
```diff
@@ -210,15 +211,30 @@ def handle_tileset(ts: minidom.Node, objs: GameObjects, builder: flatbuffers.Bui
         FlatBuffGenerated.Tileinfo.TileinfoAddGid(builder, gid)
         objs.tileinfo.append(FlatBuffGenerated.Tileinfo.TileinfoEnd(builder))
 
+def handle_tile_layer(layer: minidom.Node, builder: flatbuffers.Builder) -> str:
```
piodul: Nit: does this function really return a `str`? From looking at line 231 it seems that it returns an offset, so it should be an `int` instead.

client/dfclient/src/gameplay_state.cpp
```diff
@@ -25,6 +25,9 @@ const int FISH_FRAME_HEIGHT = 120;
 const int IMGS_PER_ROW = 10;
 const int IMGS_PER_SPECIES = 80;
 
+const int TILE_WIDTH = 128;
+const int TILE_HEIGHT = 128;
+
```
piodul: Instead of hardcoding those values, maybe we should instead encode tile with and height in the level file?

Both `<tileset>` and `<tilemap>` tags have `tilewidth` and `tileheight` attributes, so maybe we read some of those in the python script and then encode them into the level file?

@capturcus WDUT?

capturcus: Agreed with @piodul , we should probably encode tile width and height into the level file.

mparafin:
Ready to review, will be ready to merge after I add a README.md update.

deadfish.fbs
```diff
@@ -15,27 +15,31 @@ table Tileinfo {
...
+    size: Vec2;
     rotation: float;
+    isCircle: bool;
     polyverts: [Vec2];
 }
 
 table HidingSpot {
     pos: Vec2;
-    radius: float;
+    size: Vec2;
```
capturcus: why is the size property introduced in a pr regarding tiles? is it needed for the tiles?

levelpacker/levelpacker.py
```diff
@@ -55,32 +62,41 @@ def handle_objects_layer(g: minidom.Node, objs: GameObjects, builder: flatbuffer
...
+        height = 0
+        isCircle = False
+
+        if not o.getElementsByTagName('polygon'):
+            width, height = get_dim(o)
+            x += (math.cos(math.radians(rotation)) * width/2.0  - math.sin(math.radians(rotation)) * height/2.0)
+            y += (math.cos(math.radians(rotation)) * height/2.0 + math.sin(math.radians(rotation)) * width/2.0)
+            if o.getElementsByTagName('ellipse'):
+                assertCircleness(o)
+                isCircle = True
```
capturcus: i don't think any of this is related to tiles. i'll merge this in this pr but in the future please add functionalities in relevant prs

levels/tilesets/sand_tileset.tsx
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
capturcus: please do not add tilesets to the repository. after creating the ship map i think that it will be better if the tilesets are not part of the repository because they will probably differ and change much during the development and there is no point in tracking all that in the repository

capturcus (changes requested):
commiting two things at once (playerwall rework and the tiles mechanics) is not a good practice. this time i won't request changing that but please avoid doing this in the future

levels/tilesets/sand_tileset.tsx
mparafin: k

server/deadfish.cpp
```diff
@@ -87,3 +87,11 @@ CollisionMask::CollisionMask(const FlatBuffGenerated::CollisionMask* fb_Col) {
 	body->CreateFixture(&fixtureDef);
 	body->SetUserData(this);
 }
+
+Tilelayer::Tilelayer(const FlatBuffGenerated::Tilelayer* fb_Tl) : width(fb_Tl->width()), height(fb_Tl->height()),
+	    tilesize(fb_Tl->tilesize()->x(), fb_Tl->tilesize()->y()) {
+    if(!(fb_Tl->tiledata()->size())) return;
+    for(auto tile : *fb_Tl->tiledata()) {
+        this->tiledata.push_back(tile);
+    }
```
piodul: There is a `std::vector` constructor which takes a pair of iterators and copies all elements in the range ([point 5](https://en.cppreference.com/w/cpp/container/vector/vector)). Instead of writing this loop, you can do the following:

```c++
..., tiledata(fb_T1->begin(), fb_T1->end()), ...
```

This will do only one allocation (unlike the current loop) and will have the same effect (all elements from the original vector will be copied to the new one).

levelpacker/levelpacker.py
```diff
@@ -228,15 +229,48 @@ def handle_tileset(ts: minidom.Node, objs: GameObjects, builder: flatbuffers.Bui
...
+    rows = dataString.splitlines()
+    data = []
+    for row in rows:
+        for tile in row.split(','):
+            if tile == '': continue
+            data.append(int(tile))
+
+    FlatBuffGenerated.Tilelayer.TilelayerStartTiledataVector(builder, len(data))
+    for tile in data:
+        builder.PrependInt16(tile)
```
piodul, line 252: You can do:

```python
for tile in reversed(data):
    builder.PrependInt16(tile)
```

so that the final result won't be reversed.

mparafin:
@piodul Thanks for the tips <3 Applied.
Applied requested changes, fixed indicator rendering regression, ready to merge on my side.

env/dfctl
```diff
@@ -33,7 +33,7 @@ fi
 
 if [ "$1" == "test" ]
 then
-    cd test; pytest
+    cd test; pytest-3
```
capturcus: unrelated but ok

server/deadfish.cpp
```diff
@@ -86,4 +86,4 @@ CollisionMask::CollisionMask(const FlatBuffGenerated::CollisionMask* fb_Col) {
     fixtureDef.density = 1;
 	body->CreateFixture(&fixtureDef);
 	body->SetUserData(this);
-}
+}
```
capturcus: why you kill endline when endline good

capturcus: please review your changes before submitting a pr and delete meaningless changes like this

client/dfclient/src/gameplay_state.cpp
```diff
@@ -92,6 +92,32 @@ void GameplayState::LoadLevel() {
...
+	if (level->tilelayer() && level->tilelayer()->tiledata()) {
+		auto width = level->tilelayer()->width();
+		auto height = level->tilelayer()->height();
+		auto tilewidth = level->tilelayer()->tilesize()->x();
+		auto tileheight = level->tilelayer()->tilesize()->y();
+		
+		std::vector<uint16_t> tiles;
+		for(auto tile : *level->tilelayer()->tiledata()) {
+			tiles.push_back(tile);
+		}
```
capturcus: i think that creating this vector here is unnecessary, maybe you could use the tile data directly in the loops below

capturcus (changes requested):
i know this is rather nitpicky, i just wanted to bring a few things to your attention, overall i think the code is pretty good

env/dfctl
mparafin: I was supposed to be a local change, must have added it to repo accidentally. Sry
```suggestion
    cd test; pytest
```

server/deadfish.cpp
mparafin: No idea, probably been drunk or sth. That's weird af

client/dfclient/src/gameplay_state.cpp
mparafin: Indeed. Done so

mparafin:
Dun

capturcus (changes requested):
The client now crashes for me when loading map. In the loading level procedure (gameplay_state.cpp:107) the spritename variable turns out to be empty. Did you run the tests before pushing the commit? It is quite possible that my build is broken now so please check whether it works on your side.

capturcus:
Confirmed that my build is broken. Will review after fixing my setup

PR #23, opened 2020-10-13
Mob appearance and disappearance smoothing

mparafin:
ADSR for mobs. xd  Tweakable if necessary.

### DIFF
--- a/client/dfclient/include/resources.hpp
+++ b/client/dfclient/include/resources.hpp
@@ -22,6 +22,7 @@ struct Resources {
 	std::unique_ptr<ncine::AudioBufferPlayer> _wilhelmSound;
 
 	std::vector<tweeny::tween<int>> _tweens;
+	std::map<const uint16_t, tweeny::tween<int>> _mobTweens;
 
 	void UpdateTweens();
 };
--- a/client/dfclient/include/resources.hpp
+++ b/client/dfclient/include/resources.hpp
@@ -3,6 +3,7 @@
 #define __RESOURCES_HPP__
 
 #include <map>
+#include <unordered_map>
 #include <memory>
 
 #include <ncine/TextNode.h>
@@ -22,7 +23,7 @@ struct Resources {
 	std::unique_ptr<ncine::AudioBufferPlayer> _wilhelmSound;
 
 	std::vector<tweeny::tween<int>> _tweens;
-	std::map<const uint16_t, tweeny::tween<int>> _mobTweens;
+	std::unordered_map<uint16_t, tweeny::tween<int>> _mobTweens;
 
 	void UpdateTweens();
 };
--- a/client/dfclient/src/gameplay_state.cpp
+++ b/client/dfclient/src/gameplay_state.cpp
@@ -78,7 +78,7 @@ std::unique_ptr<ncine::AnimatedSprite> GameplayState::CreateNewAnimSprite(ncine:
  * @param during	time it takes
  * @return 			the created tween
 */
-tweeny::tween<int>
+static tweeny::tween<int>
 CreateAlphaTransitionTween(ncine::DrawableNode* sprite, int from, int to, int during) {
 	auto tween = tweeny::from(from)
 		.to(to).during(during).onStep(
@@ -141,7 +141,7 @@ void GameplayState::LoadLevel() {
 		objectSprite->setAnchorPoint(0, 1);
 		objectSprite->setRotation(-object->rotation());
 		objectSprite->setSize(object->size()->x(), object->size()->y());
-		if(object->hspotname()->str().empty()){
+		if (object->hspotname()->str().empty()){
 			objectSprite->setLayer(OBJECTS_LAYER);
 			this->nodes.push_back(std::move(objectSprite));
 		} else {
@@ -282,8 +282,7 @@ void GameplayState::OnMessage(const std::string& data) {
 			newMob.sprite = CreateNewAnimSprite(this->cameraNode.get(), mobData->species());
 			newMob.sprite->setAlpha(1);
 			//fade in
-			_resources._mobTweens[mobData->mobID()] = std::move(CreateAlphaTransitionTween(newMob.sprite.get(), 1, 255, MOB_FADEIN_TIME));
-			newMob.isAfterimage = false;
+			_resources._mobTweens[mobData->mobID()] = CreateAlphaTransitionTween(newMob.sprite.get(), 1, 255, MOB_FADEIN_TIME);
 			this->mobs[mobData->mobID()] = std::move(newMob);
 			mobItr = this->mobs.find(mobData->mobID());
 			firstUpdate = true;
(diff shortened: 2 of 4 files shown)


client/dfclient/src/gameplay_state.cpp
```diff
@@ -269,13 +280,23 @@ void GameplayState::OnMessage(const std::string& data) {
 			// this is the first time we see this mob, create it
 			Mob newMob;
 			newMob.sprite = CreateNewAnimSprite(this->cameraNode.get(), mobData->species());
+			newMob.sprite->setAlpha(1);
+			//fade in
+			_resources._mobTweens[mobData->mobID()] = std::move(CreateAlphaTransitionTween(newMob.sprite.get(), 1, 255, MOB_FADEIN_TIME));
```
piodul: `std::move` is completely unnecessary here. It would make sense if you moved out of a variable, but in this case move assignment operator will be called even without the `std::move`.

BTW, if you would like to learn more about how moving and rvalue references work, this is an article which makes it clear (for me at least): http://thbecker.net/articles/rvalue_references/section_01.html

client/dfclient/include/resources.hpp
```diff
@@ -22,6 +22,7 @@ struct Resources {
 	std::unique_ptr<ncine::AudioBufferPlayer> _wilhelmSound;
 
 	std::vector<tweeny::tween<int>> _tweens;
+	std::map<const uint16_t, tweeny::tween<int>> _mobTweens;
```
piodul: Nits:
- Unless you need your key-value pairs to be ordered, `unordered_map` should be more efficient
- The key type doesn't need to be const

client/dfclient/src/gameplay_state.cpp
```diff
@@ -269,13 +280,23 @@ void GameplayState::OnMessage(const std::string& data) {
...
 			mobItr = this->mobs.find(mobData->mobID());
 			firstUpdate = true;
 		}
 		Mob& mob = mobItr->second;
 		mob.setupLocRot(*mobData, firstUpdate);
 		mob.seen = true;
+		if (mob.isAfterimage == true) {
+			//re-fade in
+			_resources._mobTweens[mobData->mobID()] = std::move(
+				CreateAlphaTransitionTween(mob.sprite.get(), mob.sprite->alpha(), 255, MOB_FADEIN_TIME*(1 - mob.sprite->alpha()/255)));
```
piodul: The same thing about `std::move` applies here as well

client/dfclient/src/gameplay_state.cpp
```diff
@@ -302,11 +323,20 @@ void GameplayState::OnMessage(const std::string& data) {
 	}
 	std::vector<int> deletedIDs;
 	for (auto& mob : this->mobs) {
-		if (!mob.second.seen)
-			deletedIDs.push_back(mob.first);
+		if (!mob.second.seen) { // not seen
+			if (!mob.second.isAfterimage) {	// not seen and not afterimage
+				// fade out
+				_resources._mobTweens[mob.first] = std::move(
+					CreateAlphaTransitionTween(mob.second.sprite.get(), mob.second.sprite->alpha(), 0, MOB_FADEOUT_TIME*(mob.second.sprite->alpha()/255)));
```
piodul: Ditto

client/dfclient/src/resources.cpp
```diff
@@ -7,4 +7,9 @@ void Resources::UpdateTweens()
 		if (_tweens[i].progress() == 1.f)
 			_tweens.erase(_tweens.begin() + i);
 	}
+	for (auto it=_mobTweens.begin(); it!=_mobTweens.end(); ++it) {
```
piodul: Nit: spaces around the equality and comparison operators

client/dfclient/src/gameplay_state.cpp
```diff
@@ -520,6 +552,7 @@ void GameplayState::OnMouseButtonPressed(const ncine::MouseEvent &event) {
 	if (event.isRightButton()) {
 		// kill
 		for (auto& mob : this->mobs) {
+			if(mob.second.isAfterimage) continue;
```
piodul: Nit: put space after `if`

client/dfclient/src/gameplay_state.cpp
```diff
@@ -70,12 +70,20 @@ std::unique_ptr<ncine::AnimatedSprite> GameplayState::CreateNewAnimSprite(ncine:
...
+ * Creates a tween making a sprite fade through from one alpha value to another (specified on a 0-255 scale)
+ * @param sprite	the sprite to change
+ * @param from 		the initial alpha value
+ * @param to 		the final alpha value
+ * @param during	time it takes
+ * @return 			the created tween
+*/
 tweeny::tween<int>
-CreateHidingSpotTween(ncine::DrawableNode* hspot, int from, int to, int during) {
+CreateAlphaTransitionTween(ncine::DrawableNode* sprite, int from, int to, int during) {
```
piodul, line 82: This function can be `static`, as it is not used outside this cpp file

client/dfclient/src/gameplay_state.cpp
```diff
@@ -269,13 +280,23 @@ void GameplayState::OnMessage(const std::string& data) {
 			// this is the first time we see this mob, create it
 			Mob newMob;
 			newMob.sprite = CreateNewAnimSprite(this->cameraNode.get(), mobData->species());
+			newMob.sprite->setAlpha(1);
+			//fade in
+			_resources._mobTweens[mobData->mobID()] = std::move(CreateAlphaTransitionTween(newMob.sprite.get(), 1, 255, MOB_FADEIN_TIME));
+			newMob.isAfterimage = false;
```
piodul: No need to set it here, you already do it in the fields' declaration.

piodul: Applies to some other places as well.

client/dfclient/src/resources.cpp
```diff
@@ -7,4 +7,9 @@ void Resources::UpdateTweens()
...
+		it->second.step(1);
+		if (it->second.progress() == 1.f)
+			it->second;
```
piodul: Uh, this line doesn't do anything...

client/dfclient/src/resources.cpp
piodul: Apart from that, you can just do a range for: `for (auto& p : _mobTweens) { ... }`

client/dfclient/src/gameplay_state.cpp
```diff
@@ -302,11 +323,20 @@ void GameplayState::OnMessage(const std::string& data) {
...
 	for (auto& mob : this->mobs) {
-		if (!mob.second.seen)
-			deletedIDs.push_back(mob.first);
+		if (!mob.second.seen) { // not seen
+			if (!mob.second.isAfterimage) {	// not seen and not afterimage
+				// fade out
+				_resources._mobTweens[mob.first] = std::move(
+					CreateAlphaTransitionTween(mob.second.sprite.get(), mob.second.sprite->alpha(), 0, MOB_FADEOUT_TIME*(mob.second.sprite->alpha()/255)));
+				mob.second.isAfterimage = true;
+			} else if (mob.second.sprite->alpha() == 0){ // not seen and afterimage and alpha == 0
```
piodul: Nit: space before the curly brace

client/dfclient/src/gameplay_state.cpp
```diff
@@ -269,13 +280,23 @@ void GameplayState::OnMessage(const std::string& data) {
...
+			_resources._mobTweens[mobData->mobID()] = std::move(CreateAlphaTransitionTween(newMob.sprite.get(), 1, 255, MOB_FADEIN_TIME));
+			newMob.isAfterimage = false;
 			this->mobs[mobData->mobID()] = std::move(newMob);
 			mobItr = this->mobs.find(mobData->mobID());
 			firstUpdate = true;
 		}
 		Mob& mob = mobItr->second;
 		mob.setupLocRot(*mobData, firstUpdate);
 		mob.seen = true;
+		if (mob.isAfterimage == true) {
```
piodul: Why not just: `if (mob.isAfterimage) { ... }`

capturcus:
@mparafin I have the same issue with this as with #21 (client crashes when loading map with spritename empty). Please verify if tests pass.

capturcus (approved):
after piter's suggestions lgtm

capturcus (changes requested):
take piter's suggestions under consideration and this should be ok. sorry for the double review, i don't know github that well

client/dfclient/src/gameplay_state.cpp
mparafin: Thanks, that's right. I know how it works, I just retarded here a little bit :D

mparafin:
Applied requested changes, thanks for the review

PR #24, opened 2020-10-14
death sounds

mparafin:
The sounds of death

### DIFF
--- a/client/dfclient/src/resources.cpp
+++ b/client/dfclient/src/resources.cpp
@@ -16,8 +16,8 @@ void Resources::playKillSound(float gain) {
 
 void Resources::playRandomDeathSound() {
 	auto randIndex = rand()%_sounds.size();
-	auto it = _sounds.begin();
-	for(auto i=0; i<randIndex; ++i, ++it);
+	std::map<std::string, std::unique_ptr<ncine::AudioBuffer>>::iterator it = _sounds.begin();
+	for(auto i=0; i<randIndex; ++i) ++it;
 	_deathSound = std::make_unique<ncine::AudioBufferPlayer>(it->second.get());
 	_deathSound->play();
 }
--- a/client/dfclient/src/resources.cpp
+++ b/client/dfclient/src/resources.cpp
@@ -8,3 +8,16 @@ void Resources::UpdateTweens()
 			_tweens.erase(_tweens.begin() + i);
 	}
 }
+
+void Resources::playKillSound(float gain) {
+	_killSound->setGain(gain);
+	_killSound->play();
+}
+
+void Resources::playRandomDeathSound() {
+	auto randIndex = rand()%_sounds.size();
+	auto it = _sounds.begin();
+	for(auto i=0; i<randIndex; ++i, ++it);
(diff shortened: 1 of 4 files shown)


client/dfclient/src/resources.cpp
```diff
@@ -8,3 +8,16 @@ void Resources::UpdateTweens()
...
+	_killSound->play();
+}
+
+void Resources::playRandomDeathSound() {
+	auto randIndex = rand()%_sounds.size();
+	auto it = _sounds.begin();
+	for(auto i=0; i<randIndex; ++i, ++it);
+	_deathSound = std::make_unique<ncine::AudioBufferPlayer>(it->second.get());
+	_deathSound->play();
+}
```
capturcus: Nit: newlines at the end of file are good, please use them

client/dfclient/src/state_manager.cpp
```diff
@@ -13,22 +13,37 @@
...
+		auto absPath = rootPath.data() + std::string(SOUNDS_PATH) + "/" + std::string(file);
+		if (ncine::FileSystem::isFile(absPath.c_str())) {
+			std::cout << "loading sound " << file << " " << absPath << "\n";
+			_resources._sounds[file] = std::make_unique<ncine::AudioBuffer>(absPath.c_str());
+		}
+		file = soundDir.readNext();
+	}
+	_resources._killSoundBuffer = std::move(_resources._sounds["amongus-kill.wav"]);
+	_resources._sounds.erase("amongus-kill.wav");
+	_resources._killSound = std::make_unique<ncine::AudioBufferPlayer>(_resources._killSoundBuffer.get());
```
capturcus, line 45: a little confusing but ok

client/dfclient/src/resources.cpp
```diff
@@ -8,3 +8,16 @@ void Resources::UpdateTweens()
...
+
+void Resources::playKillSound(float gain) {
+	_killSound->setGain(gain);
+	_killSound->play();
+}
+
+void Resources::playRandomDeathSound() {
+	auto randIndex = rand()%_sounds.size();
+	auto it = _sounds.begin();
+	for(auto i=0; i<randIndex; ++i, ++it);
```
capturcus: the wasm build fails on this line with some error regarding instantiating a template. the native build works even though this doesn't look right to me

capturcus (changes requested):

client/dfclient/src/state_manager.cpp
mparafin: A bit, but imo still better than _not_ loading this particular sound effect in the earlier loop and loading it separately in the code that immediately follows. This will probably change anyway, when different species kill animations get their own sounds.

client/dfclient/src/resources.cpp
mparafin: Fixed by adding `#include <string>` to [resources.hpp](client/dfclient/include/resources.hpp). Still, took `++it` out of the braces for readability.

mparafin:
Applied requested changes.

PR #25, opened 2020-10-15
Upgrade boost from 1.65.1 to 1.74.0, change websocketpp to boost::beast

capturcus:
This PR upgrades the used Boost library from 1.65.1 to 1.74.0 (as the former doesn't include the Beast library) and changes the used websocket engine from websocketpp (that depends on the old version of boost) to the boost::beast library. This is to:
1. not rely on old version of boost (websocketpp doesn't work with any newer versions)
2. upgrade the boost library overall
3. possibly enable a windows build in the future (the package manager that I use only has the new boost in it)
4. introduce more problems that are inevitably going to arise along with all the boost black magic fuckery
5. i just kinda wanted to do it

Please be gentle, this is my first PR.

Please note that the build system has changed. You will have to download (or build) a new docker image.

### DIFF
--- a/client/dfclient/src/websocket.cpp
+++ b/client/dfclient/src/websocket.cpp
@@ -90,79 +90,103 @@ bool WebSocketEmscripten::Send(std::string& data) {
 
 #else // __EMSCRIPTEN__
 
-#include <websocketpp/config/asio_no_tls_client.hpp>
-#include <websocketpp/client.hpp>
-
-typedef websocketpp::client<websocketpp::config::asio_client> WebSocketPPClient;
-
-class WebSocketPP
+class WebSocketBeast
 	: public WebSocket
 {
 public:
 	int Connect(std::string& address) override;
 	bool Send(std::string& data) override;
-	virtual ~WebSocketPP() {}
+	virtual ~WebSocketBeast() {}
 
-	WebSocketPPClient c;
-	websocketpp::connection_hdl myHdl;
 };
 
-typedef WebSocketPP WebSocketType;
+typedef WebSocketBeast WebSocketType;
 
-std::map<void*, WebSocketPP*> clients;
+#include <boost/beast/core.hpp>
+#include <boost/beast/websocket.hpp>
+#include <boost/asio/connect.hpp>
+#include <boost/asio/ip/tcp.hpp>
+#include <cstdlib>
+#include <iostream>
+#include <string>
+#include <thread>
 
-void WebSocketPPOnMessage(websocketpp::connection_hdl hdl, WebSocketPPClient::message_ptr msg) {
-	WebSocketPP* wspp = clients[hdl.lock().get()];
-	std::lock_guard<std::mutex> guard(wspp->mq_mutex);
-	wspp->messageQueue.push_back(msg->get_payload());
-}
+namespace beast = boost::beast;         // from <boost/beast.hpp>
+namespace http = beast::http;           // from <boost/beast/http.hpp>
+namespace websocket = beast::websocket; // from <boost/beast/websocket.hpp>
+namespace net = boost::asio;            // from <boost/asio.hpp>
+using tcp = boost::asio::ip::tcp;       // from <boost/asio/ip/tcp.hpp>
 
-void WebSocketPPOnOpen(websocketpp::connection_hdl hdl) {
-	WebSocketPP* wspp = clients[hdl.lock().get()];
-	std::lock_guard<std::mutex> guard(wspp->mq_mutex);
-	wspp->toBeOpened = true;
-}
+net::io_context ioc;
+websocket::stream<tcp::socket> ws{ioc};
+beast::flat_buffer buffer;
 
-int WebSocketPP::Connect(std::string& address) {
-	// Set logging to be pretty verbose (everything except message payloads)
-	c.set_access_channels(websocketpp::log::alevel::none);
-	c.set_error_channels(websocketpp::log::elevel::all);
+void DoRead();
 
-	// Initialize ASIO
-	c.init_asio();
+void OnRead(beast::error_code ec, std::size_t bytes_transferred) {
+	if(ec == websocket::error::closed)
+		return;
 
-	// Register our message handler
-	c.set_message_handler(&WebSocketPPOnMessage);
-	c.set_open_handler(&WebSocketPPOnOpen);
-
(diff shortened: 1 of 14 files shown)


server/deadfish.hpp
```diff
@@ -71,7 +71,6 @@ struct Mob : public Collideable {
 
 struct Player : public Mob {
 	std::string name;
-	websocketpp::connection_hdl conn_hdl;
```
piodul, line 74: Was moving the connection handle to the `Mob` class intentional?

server/websocket.cpp
```diff
@@ -0,0 +1,213 @@
...
+namespace net = boost::asio;            // from <boost/asio.hpp>
+
+#include "websocket.hpp"
+#include "deadfish.hpp"
+
+class DfWebsocket;
+
+net::io_context ioc{1};
+tcp::acceptor acceptor_{ioc};
+uint16_t lastSocketID = 0;
```
piodul, line 29: Nit: I think it should be an `int`

piodul: Ok, nvm, I see that it is used to initialize an `uint16_t` field.

server/websocket.cpp
```diff
@@ -0,0 +1,213 @@
...
+class DfWebsocket : public std::enable_shared_from_this<DfWebsocket> {
+    websocket::stream<beast::tcp_stream> ws_;
+    beast::flat_buffer buffer_;
+
+public:
+    uint16_t socketID_;
+
+    // Take ownership of the socket
+    explicit
+    DfWebsocket(tcp::socket&& socket, int socketID)
```
piodul: `socketID` should have `uint16_t` type

server/websocket.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+void dfws::Run(unsigned short port)
+{
+    auto const address = net::ip::make_address("0.0.0.0");
+    beast::error_code ec;
+    tcp::endpoint endpoint{address, port};
+
+    // Open the acceptor
+    acceptor_.open(endpoint.protocol(), ec);
+    if(ec)
```
piodul: Nit: please ensure that there is a space after each `if`

server/websocket.cpp
```diff
@@ -0,0 +1,213 @@
...
+namespace websocket = beast::websocket; // from <boost/beast/websocket.hpp>
+namespace net = boost::asio;            // from <boost/asio.hpp>
+
+#include "websocket.hpp"
+#include "deadfish.hpp"
+
+class DfWebsocket;
+
+net::io_context ioc{1};
+tcp::acceptor acceptor_{ioc};
```
piodul: I think that the name of this global variable shouldn't end with an underscore. This convention should be reserved for class members.

client/dfclient/src/websocket.cpp
```diff
@@ -90,79 +90,103 @@ bool WebSocketEmscripten::Send(std::string& data) {
...
+	tcp::resolver resolver{ioc};
 
-	// Initialize ASIO
-	c.init_asio();
+	ws.binary(true);
 
-	// Register our message handler
-	c.set_message_handler(&WebSocketPPOnMessage);
-	c.set_open_handler(&WebSocketPPOnOpen);
+	auto address = fullAddress.substr(5, fullAddress.size());
```
piodul: Why do we drop first 5 characters from the beginning?? I don't understand

server/deadfish.hpp
capturcus: It was not. It's pretty incredible that it compiled and worked correctly even though this was a mistake :P

server/websocket.cpp
capturcus: Added spaces after ifs in the whole codebase.

client/dfclient/src/websocket.cpp
capturcus: edited to clarify

client/dfclient/src/websocket.cpp
```diff
@@ -90,79 +90,103 @@ bool WebSocketEmscripten::Send(std::string& data) {
...
+	WebSocketBeast* wsb = (WebSocketBeast*) webSocketManager._ws.get();
+	std::lock_guard<std::mutex> guard(wsb->mq_mutex);
+	wsb->messageQueue.push_back(str);
+	buffer.consume(buffer.size());
+	DoRead();
+}
 
-	clients[this->myHdl.lock().get()] = this;
+void DoRead() {
+	ws.async_read(buffer, &OnRead);
```
mparafin, line 148: Am I getting something wrong, or is it infinite recursion? Is it the official way to do this, isn't it something to avoid?

server/main.cpp
```diff
@@ -48,20 +48,18 @@ void addNewPlayer(const std::string &name, websocketpp::connection_hdl hdl)
...
 	new std::thread(gameThread); // leak the shit out of it yooo
 }
 
-void on_message(websocketpp::connection_hdl hdl, const server::message_ptr& msg)
+void mainOnMessage(dfws::Handle hdl, const std::string& payload)
 {
-	std::cout << "on_message\n";
-	std::cout << "message from " << (uint64_t)hdl.lock().get() << "\n";
+	std::cout << "mainOnMessage\n";
+	std::cout << "message from " << hdl << "\n";
```
mparafin: Are these left intentionally or debug?

server/websocket.cpp
```diff
@@ -0,0 +1,213 @@
...
+                return;
+            }
+            fail(ec, "read");
+        }
+
+        auto str = beast::buffers_to_string(buffer_.data());
+        onMessageHandler(socketID_, str);
+        buffer_.consume(buffer_.size());
+
+        DoRead();
```
mparafin, line 137: Same

client/dfclient/src/websocket.cpp
capturcus: This just sets up an asynchronous operation. The call returns immediately and OnRead will be called back when some data will be available to read.

server/main.cpp
capturcus: Actually both :P these messages help me to debug stuff sometimes. However they are not crucial, they are just for me. Removed.

server/websocket.cpp
capturcus: This call is synchronous, however the other one isn't so there is no infinite loop.

PR #27, opened 2020-10-23
spritepacker rework

capturcus:


spritepacker/spritepacker.py
```diff
@@ -1,63 +1,44 @@
...
+
+parser = argparse.ArgumentParser(description='Packs a folder or a series of folders into a sprite sheet.')
+parser.add_argument("--imgsinrow", help="how many images to fit in a row", required=True)
+parser.add_argument("--imgsize", help="the size of a single sprite in the target spritesheet", required=True)
+parser.add_argument("--out", help="path to the resulting spritesheet", required=True)
+parser.add_argument("--paths", help="folders to pack into the spritesheet", required=True, nargs="+")
+
+args = parser.parse_args()
+
+print(args.imgsinrow, args.imgsize, args.out, args.paths)
```
mparafin: This looks like unnecessary debug, the same is in the above line (command) when run from terminal. When run from script though, it's seemingly random. I think this should either be removed or changed into something a liitle more descriptive.

spritepacker/spritepacker.py
```diff
@@ -1,63 +1,44 @@
...
+
+import os, argparse
+from PIL import Image
+from resizeimage import resizeimage
+
+parser = argparse.ArgumentParser(description='Packs a folder or a series of folders into a sprite sheet.')
+parser.add_argument("--imgsinrow", help="how many images to fit in a row", required=True)
+parser.add_argument("--imgsize", help="the size of a single sprite in the target spritesheet", required=True)
+parser.add_argument("--out", help="path to the resulting spritesheet", required=True)
+parser.add_argument("--paths", help="folders to pack into the spritesheet", required=True, nargs="+")
```
mparafin: Nit: some defaults would be nice :P The command gets long with all these args required, and I don't seem to understand why the number of images in a row cannot be defaulted to 8 or some other neat number. Image size could probably also default to fish size for example or, better still, be read from the first source image by default.

spritepacker/spritepacker.py
```diff
@@ -1,63 +1,44 @@
...
+            files.append(f)
+    for f in sorted(files):
+        print(f)
+        ff = open(fol+"/"+f, 'r+b')
+        image = Image.open(ff)
+        if image.size[0] != image.size[1]:
+            print("image not square:", image.size, f)
+        imgs.append(image)
+
+rows = int((len(imgs)/imgsinrow))+1
```
mparafin: This produces an unnecessary empty row at the bottom of output in case of `len(imgs) % imgsinrow == 0`.

spritepacker/spritepacker.py
```diff
@@ -1,63 +1,44 @@
...
+
+imgsize = int(args.imgsize)
+imgsinrow = int(args.imgsinrow)
+
+imgs = []
+
+for fol in args.paths:
+    files = []
+    for f in os.listdir(fol):
+        if f.endswith(".png"):
```
mparafin, line 24: I think that supported formats (even (especially) if it's only PNG) should be specified somewhere outside the code. The help text looks like a nice place, and/or a descriptive exception raised when trying to feed an unsupported format, so that a user doesn't have to read into code just to know what's wrong and how to make it work.

mparafin (approved):
All of the comments are pretty much nitpicks, the tool is usable and useful, which is nice :) Do as many changes as many you agree with and merge away.
On a side note, the goldfish sprite wiggled its way into this branch, I don't think it should be here.

spritepacker/pack_all_fish.py
```diff
@@ -0,0 +1,28 @@
...
+        return "a"
+    elif key == "run":
+        return "b"
+    elif key == "attack":
+        return "c"
+    raise "wrong folder name"
+
+# sort folders so that they are in the order accepted by the client
+folders = sorted(folders, key=lambda name: name.split(".")[0]+mapAnimKeys(name.split(".")[1]))
+folders = list(map(lambda x: FISHPATH+"/"+x, folders))
```
piodul: Slightly more pythonic:

```python
folders = [FISHPATH + "/" + x for x in folders]
```

spritepacker/spritepacker.py
```diff
@@ -1,63 +1,44 @@
...
+        image = Image.open(ff)
+        if image.size[0] != image.size[1]:
+            print("image not square:", image.size, f)
+        imgs.append(image)
+
+rows = int((len(imgs)/imgsinrow))+1
+finalimg = Image.new("RGBA", (imgsinrow * imgsize, rows * imgsize))
+print("finalimg size", finalimg.size)
+
+for i in range(len(imgs)):
```
piodul: How about:

```python
for i, img in enumerate(imgs):
    # ...
```

spritepacker/spritepacker.py
capturcus: Added defaults for imgsinrow and imgsize

spritepacker/spritepacker.py
capturcus: added a check for the mentioned condition

spritepacker/spritepacker.py
capturcus: Mentioned that only PNGs are supported in the help text. Other formats should be easily supportable as well because the python library supports them but this would have to be checked separately.

PR #29, opened 2020-10-24
general skills infrastructure + the ink bomb skill

capturcus:


server/game_thread.cpp
```diff
@@ -462,13 +487,35 @@ void gameOnMessage(dfws::Handle hdl, const std::string& payload)
...
+	std::vector<int> despawns;
+	for (size_t i = 0; i < collideables.size(); i++)
+	{
+		collideables[i]->update();
+		if (collideables[i]->toBeDeleted)
+			despawns.push_back(i);
+	}
+	for (int i = despawns.size() - 1; i >= 0; i--)
+	{
+		std::cout << "erasing collideable\n";
```
mparafin: Please, let's not clutter the server log even more. Actually maybe making it optional would be a good idea but that's a whole different task.

server/skills.cpp
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	gameState.b2world->DestroyBody(this->body);
+}
+
+void InkParticle::update() {
+	this->lifetimeFrames--;
+	if (this->lifetimeFrames == 0) {
+		this->toBeDeleted = true;
+		std::cout << "ink particle toBeDeleted\n";
```
mparafin: Same as with "erasing collideable"

server/skills.cpp
```diff
@@ -0,0 +1,118 @@
...
+	gameState.inkParticles.push_back(std::move(inkPart));
+}
+
+const float INK_INIT_SPEED_BASE = 0.5;
+const float INK_INIT_SPEED_VARIABLE = 1.5;
+const int INK_COUNT = 5;
+const int INK_LIFETIME_FRAMES = 80; // 4s
+
+void executeSkillInkbomb(Player& p, UNUSED Skills skill, UNUSED b2Vec2 mousePos) {
+	std::cout << "ink bomb\n";
```
mparafin, line 40: This is ok to me, it's informative about what's going on in the game world

mparafin (changes requested):
The system works and the refactor is very nice. :heart: I noticed one problem: ink particle owner gets slowed down to walking speed when trying to run through them. Might be something with the box collision system and Mob->Player inheritance, but that's just my guess.

client/dfclient/src/gameplay_state.cpp
```diff
@@ -384,6 +380,39 @@ void GameplayState::ProcessWorldState(const void* ev) {
...
+	// todo: abstract this logic along with mobs?
+	deletedIDs.clear();
+	for (auto& p : this->inkParticles) {
+		if (!p.second.seen) {
+			deletedIDs.push_back(p.first);
+		}
+	}
+	for (auto i : deletedIDs) {
+		this->inkParticles.erase(i);
+	}
```
piodul: `std::map`'s iterators are stable. You can erase in one pass like this:

```c++
for (auto it = this->inkParticles.begin(); it != this->inkParticles.end();) {
    if (!it->second.seen) {
        it = this->inkParticles.erase(it);
    } else {
        ++it;
    }
}
```

Not sure if that's more readable, though...

If we used C++20, we could use this: https://en.cppreference.com/w/cpp/container/map/erase_if

server/deadfish.cpp
```diff
@@ -4,86 +4,90 @@
...
+	} else {
+		b2PolygonShape polyShape;
+		std::vector<b2Vec2> vertices;
+		for (auto vert : *fb_Col->polyverts()) {
+			vertices.push_back(b2Vec2(vert->x(), vert->y()));
+		}
+		polyShape.Set(vertices.data(), vertices.size());
+		fixtureDef.shape = &polyShape;
+	}
+	fixtureDef.density = 1;
```
piodul, line 90: In the future, please do unrelated whitespace changes in a separate PR.

server/mobs.cpp
```diff
@@ -240,13 +260,10 @@ Mob::~Mob()
 void handleGoldfishKill(Player& killer) {
 	if (killer.skills.size() == MAX_SKILLS)
 		return;
-	uint16_t skill = rand() % (int) Skills::SKILLS_MAX;
+	// uint16_t skill = (rand() % ((int) Skills::SKILLS_MAX - 1)) + 1;
```
piodul: Dead code, please remove

server/game_thread.cpp
```diff
@@ -462,13 +482,35 @@ void gameOnMessage(dfws::Handle hdl, const std::string& payload)
...
+		collideables.erase(collideables.begin() + despawns[i]);
+	}
```
piodul: Use `std::erase_if`:

```c++
std::erase_if(collideables.begin(), collideables.end(), [] (auto& c) {
    c->update();
    return c->toBeDeleted;
});
```

server/skills.cpp
```diff
@@ -0,0 +1,112 @@
...
+		auto& m = dynamic_cast<Mob&>(other);
+		m.bombsAffecting++;
+	} catch (...) {}
+}
+
+void InkParticle::endCollision(Collideable& other) {
+	try {
+		auto& m = dynamic_cast<Mob&>(other);
+		if (m.bombsAffecting > 0)
+			m.bombsAffecting--;
```
piodul, line 93: I think this is fragile. What if:
- Player 1 started colliding with bomb of Player 2, his `bombsAffecting` counter increased to 1,
- Player 1 started colliding with his bomb, so `bombsAffecting` counter was not increased,
- Player 1 stopped colliding with his bomb - his `bombsAffecting` counter will be erroneously decreased!

Maybe ths `InkParticle` should track which mobs it affects?

server/skills.cpp
```diff
@@ -0,0 +1,112 @@
...
+			auto collideableB = (Collideable *) edge->contact->GetFixtureB()->GetBody()->GetUserData();
+
+			auto mobA = dynamic_cast<Mob *>(collideableA);
+			auto mobB = dynamic_cast<Mob *>(collideableB);
+
+			if (mobA && mobA->bombsAffecting > 0)
+				mobA->bombsAffecting--;
+
+			if (mobB && mobB->bombsAffecting > 0)
+				mobB->bombsAffecting--;
```
piodul, line 70: Same problem as in `endCollision`...

server/skills.cpp
piodul: By the way, if we decrease the `bombsAffecting` counter below 0, it means that we have a bug related to InkParticles tracking objects incorrectly, so I think we should be checked with asserts (at least in debug mode).

piodul (commented):
Left some comments.

server/game_thread.cpp
capturcus: Sensible logging levels would be a pretty good idea for debugging. For now removed.

server/skills.cpp
capturcus: I seem to be leaving a lot of debug prints on prs, sorry

client/dfclient/src/gameplay_state.cpp
capturcus: I like erasing on one pass more, thanks for the suggestion

server/game_thread.cpp
capturcus:
```
/deadfish/server/game_thread.cpp:506:7: error: no member named 'erase_if' in namespace 'std'
        std::erase_if(collideables.begin(), collideables.end(), [] (auto& c) {
        ~~~~~^
```
I think we have to upgrade to c++20 as soon as possible. For now leaving as it was, although I like your idea much better

server/skills.cpp
capturcus: I actually found this bug in the gameplay and I had to fix it (on another branch) it before demoing the game to someone. Also, I refactored the ink particle mechanics a lot and they do not track their owners anymore (which was a lame gameplay idea to begin with) so this should not occur anymore.

server/game_thread.cpp
```diff
@@ -443,6 +446,7 @@ void executeSkill(Player& p, uint8_t skillPos, b2Vec2 mousePos) {
 		std::cout << "no such skill handler " << (uint16_t) skill << "\n";
 	}
 	skillHandler(p, skill, mousePos);
+	std::cout << "skills size " << p.skills.size() << "\n";
```
mparafin, line 448: I am... not convinced that it is necessary. :P

server/skills.cpp
```diff
@@ -31,23 +31,25 @@ void launchInkParticle(Player& p, b2Vec2 direction) {
...
 		launchInkParticle(p, direction);
 	}
 }
 
 InkParticle::InkParticle(b2Body* b, Player& o) 
-	: owner(o) {
+{
 	this->body = b;
 	this->lifetimeFrames = INK_LIFETIME_FRAMES;
 }
```
mparafin: If `Player& o` argument is no longer necessary, it should probably be removed

server/game_thread.cpp
mparafin:
```
	for (auto it = collideables.begin(); it != collideables.end();) {
		(*it)->update();
		if ((*it)->toBeDeleted) {
			collideables.erase(it);
		} else {
			++it;
		}
	}
```
seems to be working here as well (version with `for(...; ...; ++it)` and `collideables.erase(it); --it;` too), but probably because the changes during testing weren't critical enough to cause vector reallocation. Can somebody (@piodul, @capturcus) verify my thinking here? :D

server/game_thread.cpp
```diff
@@ -163,8 +163,6 @@ flatbuffers::Offset<FlatBuffGenerated::Mob> createFBMob(flatbuffers::FlatBufferB
 {
 	auto distance = b2Distance(m->body->GetPosition(), player.body->GetPosition());
```
mparafin: Compiler yells that this variable is no longer necessary. If that's true, please remove :)

levelpacker/levelpacker.py
```diff
@@ -267,26 +272,37 @@ def process_level(path: str):
...
+            processedLayers += 1
         elif group_name == 'meta':
             handle_meta_layer(g, objs, builder)
+            processedLayers += 1
         else:
             print("WARNING: Unknown group {}".format(group_name))
     
+    if processedLayers < 6:
+        print("WARNING: a layer is missing")
+
```
mparafin, line 305: This might actually be a mistake on my side - I wanted to make it so that a missing layer (tile layer e.g.) would be acceptable, but it seems I have failed

mparafin (approved):
I see you've added the killDistance refactor here. It's cool, I like it, but because it's quite a big gameplay change I'm wondering if it shouldn't be on a separate branch, just for order and tidiness. But that's just philosophy, let this one be here. That's a :+1: from me

mparafin:
Oh, and also: ink cloud position is not lerped, so it looks laggy. This is something to be repaired, perhaps in another task

levelpacker/levelpacker.py
capturcus: I think that when you were writing this it actually worked as you expected. However due to later changes that I've introduced the game started to crash when it couldn't find certain layers.

PR #30, opened 2020-11-09
Blink + minor refactors

capturcus:
The blink skill itself turned out to be just 4 lines of code. However while playing and debugging I also discovered other issues. Things that are in this PR:

- blink skill
- a fix in the client: mouse coordinates were sent wrongly on skill activate
- removed owner of the ink bomb, changed mechanics to be more fair to all players
- changed storing civilians on the server from a vector to a map to achieve stable iterators and avoid segfaults in specific edge cases

### DIFF
--- a/server/mobs.cpp
+++ b/server/mobs.cpp
@@ -265,7 +265,7 @@ Mob::~Mob()
 void handleGoldfishKill(Player& killer) {
 	if (killer.skills.size() == MAX_SKILLS)
 		return;
-	uint16_t skill = (int) Skills::INK_BOMB; // all skills are ink bombs bc it's the only working skill for now
+	uint16_t skill = (uint16_t) (rand() % 2 == 0 ? Skills::INK_BOMB : Skills::BLINK);
 	killer.skills.push_back(skill);
 	killer.sendSkillBarUpdate();
 }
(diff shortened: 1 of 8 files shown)


server/mobs.cpp
```diff
@@ -268,7 +268,7 @@ Mob::~Mob()
 void handleGoldfishKill(Player& killer) {
 	if (killer.skills.size() == MAX_SKILLS)
 		return;
-	uint16_t skill = (int) Skills::INK_BOMB; // all skills are ink bombs bc it's the only working skill for now
+	uint16_t skill = (uint16_t) (rand() % 2 == 0 ? Skills::INK_BOMB : Skills::BLINK);
```
mparafin, line 271: That's still a temporary solution xD But ok

server/mobs.cpp
```diff
@@ -287,7 +287,6 @@ void Civilian::handleKill(Player& killer) {
 	flatbuffers::FlatBufferBuilder builder;
 	auto ev = FlatBuffGenerated::CreateDeathReport(builder, killer.playerID, -1);
 	sendServerMessage(killer, builder, FlatBuffGenerated::ServerMessageUnion_DeathReport, ev.Union());
-	std::cout << "sent a death report\n";
```
mparafin, line 290: :heart:

mparafin (approved):
Codewise - _I find no guilt in this PR_, clean and straightforward.
Gameplaywise - blink range is definitely a thing to balance (it's way too much afaic) and the first thing I've done in testing is blinked myself out of the map xD It's an issue to be adressed probably on the map side though, but just pointing out that it exists.

mparafin:
Another "problem": lerping mechanism makes the player travel really quickly to the destination point instead of really blinking. Not much of an issue, unless this makes the fish seemingly go over (or through) obstacles - also it may be difficult to add extra visual effects (animations/particles).

PR #31, opened 2020-11-09
Mob manipulators

capturcus:
- mob manipulator skills (dispersor and attractor)
- presimulate 1000 frames so that the map isn't empty at the beginning of the game
- validate ink bomb visibility with line of sight
- delete Skills::SKILL_NONE as it was unused

### DIFF
--- a/client/dfclient/src/gameplay_state.cpp
+++ b/client/dfclient/src/gameplay_state.cpp
@@ -419,7 +419,6 @@ void GameplayState::ProcessWorldState(const void* ev) {
 		auto sprite = std::make_unique<ncine::Sprite>(this->cameraNode.get(), _resources.textures[manipTexture].get());
 		sprite->setLayer((unsigned short) Layers::MOB_MANIPULATORS);
 		sprite->setPosition({manipulator->pos()->x() * METERS2PIXELS, -manipulator->pos()->y() * METERS2PIXELS});
-		sprite->setScale(manipulator->dispersor() ? 120./524. : 120./288.); // xd
 		this->manipulators.push_back(std::move(sprite));
 	}
 }
--- a/client/dfclient/src/gameplay_state.cpp
+++ b/client/dfclient/src/gameplay_state.cpp
@@ -411,15 +411,15 @@ void GameplayState::ProcessWorldState(const void* ev) {
 			++it;
 	}
 
-	this->manipulators.clear();
 	// handle mob manipulators
+	this->manipulators.clear();
 	for (int i = 0; i < worldState->mobManipulators()->size(); i++) {
 		auto manipulator = worldState->mobManipulators()->Get(i);
 		auto manipTexture = manipulator->dispersor() ? "dispersor.png" : "attractor.png";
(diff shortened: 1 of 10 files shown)


client/dfclient/src/gameplay_state.cpp
```diff
@@ -409,6 +410,18 @@ void GameplayState::ProcessWorldState(const void* ev) {
...
+
+	// handle mob manipulators
+	this->manipulators.clear();
+	for (int i = 0; i < worldState->mobManipulators()->size(); i++) {
+		auto manipulator = worldState->mobManipulators()->Get(i);
+		auto manipTexture = manipulator->dispersor() ? "dispersor.png" : "attractor.png";
+		auto sprite = std::make_unique<ncine::Sprite>(this->cameraNode.get(), _resources.textures[manipTexture].get());
+		sprite->setLayer((unsigned short) Layers::MOB_MANIPULATORS);
+		sprite->setPosition({manipulator->pos()->x() * METERS2PIXELS, -manipulator->pos()->y() * METERS2PIXELS});
+		sprite->setScale(manipulator->dispersor() ? 120./524. : 120./288.); // xd
```
mparafin: xd
I suppose this is due to different image scale. I think this is something that shouldn't be in the code, but rather the images should be resized to the necessary dimensions.

server/mobs.cpp
```diff
@@ -180,6 +181,28 @@ void Civilian::collisionResolution() {
...
+			this->targetPosition = b2g(last.pos);
+		else {
+			auto toManip = this->body->GetPosition() - last.pos;
+			toManip.Normalize();
+			this->targetPosition = b2g(this->body->GetPosition() + toManip);
+		}
+		Mob::update();
+		return;
+	} else if (this->seenAManip) {
+		this->collisionResolution();
```
mparafin, line 205: This line makes many mobs disappear at random at the end of the mob manipulator effect duration.
1) It looks weird
2) If the tactical goal of MM is to simply expose players by moving every mob except for them, it's ok. But if I would like, for example, to create a dense mob cluster somewhere for a while, that's becoming impossible. If I understand the mechanics correctly, the mobs would go to the nearest waypoint anyway, so the cluster wouldn't be persistent.

mparafin (changes requested):
Works nice, looks nice :) But I think the problems in the comments need to be addressed before merge

server/mobs.cpp
capturcus: fixed

client/dfclient/src/gameplay_state.cpp
capturcus: resized the images in the data repository, removed this line

mparafin (changes requested):
---

server/deadfish.hpp
```diff
@@ -210,6 +210,11 @@ struct Level {
 	glm::vec2 size;
 };
 
+struct MobManipulator {
+	b2Vec2 pos;
+	bool dispersor;
```
piodul: How about we introduce an enum that indicates the type of a manipulator?

mparafin (approved):
:+1:

PR #32, opened 2020-11-12
Killcount mechanics

mparafin:
Game mechanics regarding multiplayer interactions, such as:
- multikill (several kills in quick succession)
- killing spree (multiple kills without dying)
- shutdown (ending sb's killing spree)
- domination (killing a player multiple times without dying from the player's hand)
- revenge (killing your ~~dominatrix~~ dominating player)
- comeback (finally getting a kill after dying multiple times)

Includes server-side mechanics, flatbuffer-side transmission and client-side sweet visual treatment.
**Doesn't** include game log with messages about events in the game (e.g. _asd for revenge on asdd!_) - this will come in the next PR.
Enjoy! ^^


client/dfclient/include/resources.hpp
```diff
@@ -17,28 +17,35 @@ static const char* TEXTURES_PATH = "textures";
...
+	/** Plays a sound
+	 * @param sound which sound type to play
 	 * @param gain volume (1.0f is 100%)
 	*/
-	void playKillSound(float gain = 1.0f);
-
-	/** Plays random death sound
-	*/
-	void playRandomDeathSound();
+	void playSound(SoundType soundType, float gain = 1.0f);
```
capturcus, line 48: Why are there sound changes on a killstreak pr? It's already huge man, have mercy

client/dfclient/include/text_creator.hpp
```diff
@@ -0,0 +1,118 @@
...
+	float _textScale;
+	int _from, _hold, _decay;
+	Layers _layer;
+
+	std::string _lastText;
+	ncine::Color _lastColor;
+	float _last_x;
+	float _last_y;
+	float _lastScale;
+	int _last_from, _last_hold, _last_decay;
```
capturcus: I appreciate the underscore naming convention

client/dfclient/src/process_death_report.cpp
```diff
@@ -0,0 +1,449 @@
...
+			theTextCreator->setTweenParams(255, 45, 90); // timing for "you killed <player>"
+			theTextCreator->setColor(0, 220, 0);
+			auto killtext = theTextCreator->CreateText(
+				"you killed " + gameData.players[deathReport->victim()].name + "!",
+				std::nullopt,
+				textsXPos[1], textsYPos[0],
+				3.0f,
+				std::nullopt, std::nullopt, std::nullopt,
+				std::nullopt,
+				tweeny::easing::sinusoidalIn
```
capturcus: it was all worth it

client/dfclient/src/state_manager.cpp
```diff
@@ -41,15 +41,26 @@ StateManager::StateManager() {
 		}
 		file = soundDir.readNext();
 	}
+
+	_resources._sounds.erase("you-failed.wav"); // temporary disable, it will be useful as a gameover sound effect
```
capturcus: maybe just remove it from the data repository, I don't see much use for this sound effect (and also it's copyrighted af)

server/mobs.cpp
```diff
@@ -291,6 +297,72 @@ Mob::~Mob()
...
+struct MultiplayerMechanicsInfo {
+	uint16_t multikill = 0;
+	uint16_t killing_spree = 0;
+	uint16_t shutdown = 0;
+	uint16_t domination = 0;
+	uint16_t revenge = 0;
+	uint16_t comeback = 0;
+};
+
+MultiplayerMechanicsInfo handleMultiplayerMechanics(Player& killer, Player& victim) {
```
capturcus: "handle multiplayer mechanics" could mean anything in a multiplayer game, maybe we could rename this to something more descriptive like for instance `handleKillstreakMechanics` or `handleKillStatistics`

server/mobs.cpp
```diff
@@ -291,6 +297,72 @@ Mob::~Mob()
 	}
 }
 
+struct MultiplayerMechanicsInfo {
```
capturcus: I think that "multiplayer mechanics" isn't really descriptive, maybe we could change it to "killstreak mechanics"? what do you think?

server/mobs.cpp
```diff
@@ -291,6 +297,72 @@ Mob::~Mob()
...
+	MultiplayerMechanicsInfo ret;
+
+	if (killer.multikillTimer && killer.multikillCounter > 1) {
+		killer.points += killer.multikillCounter * MULTIKILL_REWARD;
+		ret.multikill = killer.multikillCounter;
+	}
+
+	// killing spree
+	if (killer.killingSpreeCounter >= KILLING_SPREE_THRESHOLD) {
+		std::cout << killer.name << " KILLING SPREE " << killer.killingSpreeCounter << std::endl;
```
capturcus, line 328: I like that the server itself gets emotional over killing sprees

client/dfclient/src/process_death_report.cpp
```diff
@@ -0,0 +1,449 @@
+/**
```
capturcus: At first it was unclear to me that this is in fact part of the GameplayState class. Maybe we could rename this file to `gameplay_state_deaths.cpp` or something like this?

capturcus (changes requested):
Overall very good work, I'm impressed. I just had a few nitpicks

client/dfclient/include/resources.hpp
mparafin: xD Sry. My thought stream: I'm doing kill management -> goldfish kills are kills to be managed -> playing sounds is part of kill management -> doing a quick sound refactor was about as much work as sticking to the previous structure :D But you're right, it should go in a separate PR, like font_upgrade did (similar thing)

client/dfclient/src/process_death_report.cpp
mparafin: duh :P

client/dfclient/src/state_manager.cpp
mparafin: I thought of it as a thing that could play when you finish last. Copyright is a point, but... nearly every single sound we use in this game is copyrighted xDD All I'm counting on is the meme-ish-ness of the game when it comes to sound. But we can remove this if you want

server/mobs.cpp
mparafin: The best I managed to come up with is "killcount mechanics", because that's basically what it is about, counting kills. If you're ok with the name, I'll push an update

client/dfclient/src/process_death_report.cpp
mparafin: Is `gameplay_state_death_reports.cpp` ok?

client/dfclient/src/state_manager.cpp
capturcus: Copyright point taken, all of the sounds are copyrighted, so this doesn't matter. However I don't want to be punishing players that were last, it would probably be better to award the ones that were first instead. Please remove

server/mobs.cpp
capturcus: ok

client/dfclient/src/process_death_report.cpp
capturcus: yes

mparafin:
Changes applied as requested

client/dfclient/include/state_manager.hpp
```diff
@@ -19,6 +20,8 @@ struct StateManager {
 	void OnMouseButtonPressed(const ncine::MouseEvent &event);
 	void OnMouseMoved(const ncine::MouseState &state);
 
+	TextCreator CreateTextCreator();
```
piodul, line 23: So, `StateManager` is essentially a `TextCreatorCreator`? xd

client/dfclient/include/text_creator.hpp
```diff
@@ -0,0 +1,118 @@
...
+		std::optional<ncine::Color> color = std::nullopt,
+		std::optional<float> pos_x = std::nullopt,
+		std::optional<float> pos_y = std::nullopt,
+		std::optional<float> scale = std::nullopt,
+		std::optional<int> from = std::nullopt,
+		std::optional<int> hold = std::nullopt,
+		std::optional<int> decay = std::nullopt,
+		std::optional<Layers> layer = std::nullopt,
+		const T& easing = tweeny::easing::linear
+		);
```
piodul, line 96: I see what you are trying to do here with the optional parameters, but... methods with such a large number of arguments are unwieldy to use. What if I want to specify a custom `layer`, and leave all other parameters as defaults? I would have to specify nullopts for all previous parameters. Moreover, if I miss one of the preceding arguments, it might turn out that I accidentally assigned the `decay` parameter. Same problems will happen if we decide to add another argument somewhere in the middle.

How about this: if somebody wishes to create a text, they create their own copy of the creator, set some parameters and then create the text?

client/dfclient/include/text_creator.hpp
piodul: The underscore naming convention is reserved for _private_ fields only. In this case, it seems like the convention used is correct, but the actual visibility is not - I think those fields should be private.

client/dfclient/include/gameplay_state.hpp
```diff
@@ -89,8 +93,19 @@ struct GameplayState
 
 	void updateRemainingText(uint64_t remainingFrames);
 
+	void processKill(const FlatBuffGenerated::DeathReport* deathReport);
+	void processDeath(const FlatBuffGenerated::DeathReport* deathReport);
+	void processObituary(const FlatBuffGenerated::DeathReport* deathReport);
+	std::unique_ptr<ncine::TextNode> processMultikill(int multikillness);
```
piodul: It's... a little strange that this method returns a text node. Looking at the signature alone, it's not really clear what the value being returned is supposed to mean. Please at least add a comment which tells what is the meaning of the return value.

client/dfclient/include/resources.hpp
```diff
@@ -17,28 +17,35 @@ static const char* TEXTURES_PATH = "textures";
 static const char* LEVELS_PATH = "../levels";
 static const char* SOUNDS_PATH = "sounds";
 
+enum SoundType {
```
piodul: `enum class`

client/dfclient/include/text_creator.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+private:
+	template <typename T>
+	std::unique_ptr<ncine::TextNode> doCreateText(std::string text, ncine::Color color,
+		float pos_x, float pos_y, float scale, int from, int hold, int decay, Layers layer, const T& easing, bool isOutline);
+};
+
+#include "text_creator_templates.hpp"
+
+extern std::optional<TextCreator> textCreator;
```
piodul: Do we really need a global...?

client/dfclient/include/text_creator_templates.hpp
```diff
@@ -0,0 +1,92 @@
...
+	if (!isOutline) {
+		_lastText = text;
+		_lastColor = color;
+		_last_x = pos_x;
+		_last_y = pos_y;
+		_lastScale = scale;
+		_last_from = from;
+		_last_hold = hold;
+		_last_decay = decay;
+	}
```
piodul, line 90: Are those variables necessary? Why can't we just use a parameter X while creating outline instead of using its `_last_X` version? I think it makes it more complicated than it needs to be.

client/dfclient/include/text_creator.hpp
```diff
@@ -0,0 +1,118 @@
...
+/** Simplifies creating various in-game texts, such as UI messages on player kills.
+ * Designed to allow long-term and one-off text parameter customization. As such,
+ * stores the default values, which are customizable directly or through setter methods
+ * and are supplied if no value for a parameter is given.
+ * 
+ * To simplify creating outlines, this manager also stores the values used to create
+ * the last text. The values are supplied automatically to the outline-creating method,
+ * with the exception of tween easing, which has to be given once more as a parameter.
+ */
+class TextCreator {
```
piodul, line 17: This is a builder pattern, so maybe the `TextBuilder` name would be a little more fitting?

client/dfclient/include/text_creator_templates.hpp
```diff
@@ -0,0 +1,92 @@
...
+		Layers::TEXT_OUTLINES,
+		easing,
+		true
+	);
+}
+
+
+template <typename T>
+std::unique_ptr<ncine::TextNode> TextCreator::doCreateText(std::string text, ncine::Color color,
+        float pos_x, float pos_y, float scale, int from, int hold, int decay, Layers layer, const T& easing, bool isOutline) {
```
piodul, line 67: If we get rid of `_last_X` parameters, then most of the arguments here become obsolete.

client/dfclient/src/gameplay_state_death_reports.cpp
```diff
@@ -0,0 +1,449 @@
...
+#include "resources.hpp"
+#include "text_creator.hpp"
+
+// translation-unit-local variables used throughout the code in this file
+static std::optional<TextCreator> theTextCreator;	// instanced here for potential future thread-safety
+static ncine::SceneNode& rootNode = ncine::theApplication().rootNode();
+static float screenWidth;
+static float screenHeight;
+static std::array<float, 3> textsXPos;
+static std::array<float, 4> textsYPos;
```
piodul: No, please, let's not add globals if it's not necessary.

If you need to keep some state across several function calls, why not wrap those globals into a "context" structure, or create a new class for the sole purpose of processing the death report?

client/dfclient/include/text_creator.hpp
```diff
@@ -0,0 +1,118 @@
...
+	// existent for code clarity
+	inline void setColor(ncine::Color color) { _textColor = color; }
+	inline void setColor(int r, int g, int b, int a = 255) { _textColor = ncine::Color(r,g,b,a); }
+
+	inline void setPosition(float x, float y) { _pos_x = x; _pos_y = y; }
+	inline void setScale(float scale) { _textScale = scale; }
+	inline void setFrom(int from) { _from = from; }
+	inline void setHold(int hold) { _hold = hold; }
+	inline void setDecay(int decay) { _decay = decay; }
+	inline void setTweenParams(int from, int hold, int decay) { setFrom(from); setHold(hold); setDecay(decay); }
```
piodul, line 48: The name of this function is really nondescript. IMO it's better to call those methods separately. Or, rename it to `setFromHoldDelay`?

client/dfclient/src/gameplay_state.cpp
```diff
@@ -625,6 +568,14 @@ StateType GameplayState::Update(Messages m) {
 		closestMob->hoverMarker->setLayer((unsigned short)Layers::INDICATOR);
 	}
 
+	// clean up transparent text
+	for (auto it = this->textNodes.begin(); it != this->textNodes.end(); ++it) {
+		if ((*it)->alpha() == 0) {
+			textNodes.erase(it);
+			--it;
+		}
+	}
```
piodul: A more performant version:

```c++
auto new_end = std::remove_if(this->textNodes.begin(), this->textNodes.end(), [] (auto& p) {
    return p->alpha() == 0;
});
this->textNodes.resize(std::distance(this->textNodes.begin(), new_end));
```

client/dfclient/src/gameplay_state_death_reports.cpp
```diff
@@ -0,0 +1,449 @@
...
+			theTextCreator->setScale(1.2f);
+			theTextCreator->setColor(0, 200, 0);
+			texts.push_back(theTextCreator->CreateText("enemy killed"));
+			texts.push_back(theTextCreator->CreateOutline());
+
+			texts.push_back(theTextCreator->CreateText(std::to_string(KILL_REWARD)));
+			texts.push_back(theTextCreator->CreateOutline());
+
+			int totalScore = KILL_REWARD;
+			int scoreBonus;
```
piodul: Let's avoid uninitialized variables. Set it to `0` here, or even better, declare it in each `if` branch on demand.

client/dfclient/src/gameplay_state_death_reports.cpp
```diff
@@ -0,0 +1,449 @@
...
+					.to(totalScore).during(20).onStep(
+						[scoreNode, scoreOutlineNode] (tweeny::tween<int>& t, int v) -> bool {
+							scoreNode->setString(nctl::String(std::to_string(v).c_str()));
+							scoreOutlineNode->setString(nctl::String(std::to_string(v).c_str()));
+							return false;
+						}
+					));
+		}
+		for (auto& text : texts) {
+		this->textNodes.push_back(std::move(text));
```
piodul: Missing indent.

client/dfclient/src/text_creator.cpp
```diff
@@ -0,0 +1,12 @@
...
+
+#include "gameplay_state.hpp"
+
+TextCreator::TextCreator(Resources& resources) : _resources(resources) {
+    _textColor = ncine::Color(0, 0, 0);
+    _pos_x = 0;
+    _pos_y = 0;
+    _textScale = 1.0f;
+    _from = 255; _hold = 60; _decay = 60;
+    _layer = Layers::TEXT;
```
piodul: You can assign such defaults in class body, right after field declaration.

E.g.:

```c++
class X {
public:
    ncine::Color color = ncine::Color(1, 2, 3);
};
```

capturcus (approved):
after taking into consideration piter's suggestions it will be ok to me

capturcus (changes requested):
after piter's suggestions it will be ok

client/dfclient/include/gameplay_state.hpp
mparafin: You're right, will do so

client/dfclient/src/gameplay_state_death_reports.cpp
mparafin: That's right, sorry

client/dfclient/include/text_creator.hpp
mparafin: We actually do not. I thought it would be a good idea, but after consideration it is not. I will remove this.

client/dfclient/src/text_creator.cpp
mparafin: This is due to necessity to initialize _resources here, so I figured that all the values should be initialized in the same place. Probably not the best idea though, as it makes it less clear what the default initialized values are, when they are in another file. Will move init to .hpp.

client/dfclient/src/gameplay_state_death_reports.cpp
mparafin: As far as I can tell, these variables are global and visible only in the scope of this translation unit (for example, I cannot use them in `gameplay_state.cpp`, which is in the same place in the file dependency hierarchy). Is this really so bad? These are just helpers, universal variables that I use in every function in this file. Making them global allows me to use them in these functions without having to declare them each time. If there is some more evil coming from them being global, please tell me, so I can learn :D But I can't see anything wrong here.

Ideally they would be `const`, but they cannot be initialized at the start of the program, because their values (screen width, for example) are not specified yet. If you have a better idea how to make them visible in all the functions (and `const`), please tell. :D (I assume `#define` is not the recommended option :P)

client/dfclient/include/text_creator.hpp
mparafin: For me, this name is more descriptive than `setFromHoldDelay`, because it tells what the values are actually used for :P It requires knowledge of the TextCreator's workings, but not more than it is required to use it, I feel. If one prefers to use the separate methods, well - that's why they are provided :P

client/dfclient/include/text_creator_templates.hpp
mparafin: Same as in https://github.com/capturcus/deadfish/pull/32#discussion_r538373179

client/dfclient/include/text_creator.hpp
mparafin: I think it's just a flavor :) I felt the TextCreator was just a cooler name, but if you insist it can be changed. I'm just kinda attached to it emotionally, so you need to state it clearly that you want it changed xD :sweat_smile:

client/dfclient/include/text_creator_templates.hpp
mparafin: It's because of the `createText` parameters :D As long as they exist, we can't rely on the `X` variables to be the values used during creating the last text. Addressed in another conversation (https://github.com/capturcus/deadfish/pull/32#discussion_r538406315).

client/dfclient/include/text_creator.hpp
mparafin: Well, I wanted to make this tool as universal as can be. If you want to, you can do exactly as you have described above. I just think this:
```
textCreator->CreateText(
	"you ended " + gameData.players[deathReport->victim()].name + "'s killing spree!",
	ncine::Color(255, 70, 0),
	textsXPos[1], textsYPos[2],
	1.5f,
	255, 60, 120
);
```
is more clear and readable than this:
```
TextCreator textCreator;
textCreator.setColor(ncine::Color(255, 70, 0));
textCreator.setPosition(textsXPos[1], textsYPos[2]);
textCreator.setScale(1.5f);
textCreator.setTweenParams(255, 60, 120);
textCreator->CreateText("you ended " + gameData.players[deathReport->victim()].name + "'s killing spree!");
```

Of course, the second version is very descriptive for someone who doesn't know the tool, but the point is, the function is used very frequently, at least in this part of code. When you get to know what the parameters are for, it becomes very clear, and all the `textCreator.setX(` part becomes obsolete and just obscures the code.

Regarding your example, you don't need to `std::nullopt` your way all the way to layer, that's what the setter functions are for. You can just do:
```
textCreator.setLayer(Layers::someLayer);
textCreator.CreateText("stuff");
```
Or any mix of the two approaches.

Having said all this, I think it should stay the way it is, especially because of the support for different use styles. But you're welcome to prove me wrong :P

client/dfclient/include/state_manager.hpp
mparafin: Kind of :D

client/dfclient/include/text_creator.hpp
piodul: I'm not insisting too much. I'm fine with leaving it as it is.

client/dfclient/src/gameplay_state_death_reports.cpp
piodul: > As far as I can tell, these variables are global and visible only in the scope of this translation unit (for example, I cannot use them in gameplay_state.cpp, which is in the same place in the file dependency hierarchy).

OK, you are right that the variables are not global - their scope is restricted to only one file. In such, case, I have two problems with them:

- They make the functions in this file not thread-safe. Not sure if a problem in this case, but it is in general. We can't really say that those variables will always be initialized to the same value, because the screen (window) size can change.
- This
  ```
  static ncine::SceneNode& rootNode = ncine::theApplication().rootNode();
  ```
  is dangerous, because it's prone to the [static initialization order fiasco](https://en.cppreference.com/w/cpp/language/siof). If you have statically-initialized variables in file A which depend in some way (either directly or indirectly, e.g. through a function) on static variables in file B, then there is a risk that variables in A will be attempted to be initialized before those in file B.

What I suggest to do is to create a class/struct which will live for the duration of processing the death report. The helper static variables can be converted into `const` members, and initialized in the constructor. Something like this:

```c++
// Somewhere in the beginning of the gameplay_state_death_reports.cpp file:
class DeathReportProcessor {
private:
    const TextCreator _textCreator;
    const ncine::SceneNode& _rootNode;
    const float _screenWidth, _screenHeight;
    const std::array<float, 3> _textsXPos;
    const std::array<float, 4> _textsYPos;

public:
    DeathReportProcessor(Resources& resources)
        : _textCreator(resources)
        , _rootNode(ncine::theApplication().rootNode())
        , _screenWidth(ncine::theApplication().width())
        , _screenHeight(ncine::theApplication().height())
        , _textsXPos{0.41f * _screenWidth, 0.5f * _screenWidth, 0.59f * _screenWidth}
        , _textsYPos{0.75f * _screenHeight, 0.5f * _screenHeight, 0.30f * _screenHeight, 0.25f * _screenHeight}
    {}

    // Main method declaration
    void ProcessDeathReport(FlatBuffGenerated::DeathReport* report);

private:
    // Convert other helper methods and declare them here
    void processKill(const FlatBuffGenerated::DeathReport* deathReport);
    std::unique_ptr<ncine::TextNode> processMultikill(int multikillness);
    void endKillingSpree();
    // et cetera
};

void GameplayState::ProcessDeathReport(void const* ev) {
    const auto deathReport = (const FlatBuffGenerated::DeathReport*)ev;
    DeathReportProcessor processor(_resources);
    processor.ProcessDeathReport(ev);
}

// Here go the definitions of DeathReportProcessor's methods...
```

This way you can still access the variables from the helper functions without having to prefix them in any way (only with underscore), and now you have a separate instance of those variables for each ProcessDeathReport, so there is no risk of accidental sharing, and no risk of static order initialization fiasco.

client/dfclient/include/text_creator.hpp
piodul: I'm fine with leaving it as it is.

piodul:
I'm OK with merging after the issue in `gameplay_state_death_reports.cpp ` is addressed.

client/dfclient/src/gameplay_state_death_reports.cpp
mparafin: Well, that seems to be our golden answer to all the problems. Thanks for all the help, will implement :)

mparafin:
Rebased onto `movable_refactor`

client/dfclient/include/text_creator_templates.hpp
```diff
@@ -0,0 +1,92 @@
...
+			hold.value_or(_hold),
+			decay.value_or(_decay),
+			layer.value_or(_layer),
+			easing,
+			false);
+	}
+
+template <typename T>
+std::unique_ptr<ncine::TextNode> TextCreator::CreateOutline(const T& easing) {
+	if(_lastColor == ncine::Color::Black) return nullptr;
```
piodul: Nit: Please adhere to our coding style and put space after the `if`

client/dfclient/src/death_report_processor.cpp
```diff
@@ -0,0 +1,433 @@
...
+	} else {
+		// somebody died
+		return processObituary(deathReport);
+	}
+}
+
+/* ----------- I KILLED SOMEBODY ------------ */
+
+DrawableNodeVector DeathReportProcessor::processKill(const FlatBuffGenerated::DeathReport* deathReport) {
+		_resources.playSound(SoundType::KILL);
```
piodul: This method is indented too much. Every line is indented with at least 2 tabs - please fix it.

client/dfclient/src/resources.cpp
```diff
@@ -2,25 +2,50 @@
...
+	case SoundType::KILL:
+		soundPlayer = _killSound.get();
+		break;
+	case SoundType::GOLDFISH:
+		soundPlayer = _goldfishSound.get();
+		break;
+	case SoundType::DEATH:
+		auto randIndex = rand()%_sounds.size();
+		auto it = _sounds.begin();
+		for(auto i=0; i<randIndex; ++i) ++it;
```
piodul: Nit: space after `for`

piodul (approved):
Left some minor comments about style, after fixing them it will LGTM.

mparafin:
> Left some minor comments about style, after fixing them it will LGTM.

Done! Ready to merge by @capturcus when [`movable_refactor`](39) is merged.

capturcus (approved):

PR #33, opened 2020-11-14
Improvements to the main Makefile

piodul:
- adds two new targets for re-running cmake in docker
- adds a flag which allows to disable using sudo for docker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,26 @@
+ifndef DEADFISH_DOCKER_NOSUDO
+dockercmd=sudo docker
+else
+dockercmd=docker
+endif
+
 all: server client-wasm client-native
-.PHONY: server client-wasm client-native test
+.PHONY: server client-wasm client-native test recmake-dfclient-wasm recmake-dfclient-native
 
 server:
-	sudo docker exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/server/build && make -j5'
+	$(dockercmd) exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/server/build && make -j5'
 
 client-wasm:
-	sudo docker exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/dfclient-wasm && make -j5'
+	$(dockercmd) exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/dfclient-wasm && make -j5'
 
 client-native:
-	sudo docker exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/dfclient-native && make -j5'
+	$(dockercmd) exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/dfclient-native && make -j5'
+
+recmake-dfclient-wasm:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,14 +1,26 @@
...
 
 client-native:
-	sudo docker exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/dfclient-native && make -j5'
+	$(dockercmd) exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/dfclient-native && make -j5'
+
+recmake-dfclient-wasm:
+	$(dockercmd) exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/ && ./recmake_dfclient_wasm.sh'
+
+recmake-dfclient-native:
+	$(dockercmd) exec --user ubuntu -it deadfish-env /bin/bash -c 'cd /deadfish/client/ && ./recmake_dfclient_native.sh'
```
mparafin, line 23: This is basically duplicated functionality from `dfctl`. I think we should settle on one tool and commit to using and developing it, instead of several different ones. Do you think that the no-sudo docker thing can be implemented in `dfctl`, or should we switch to using `make`?

Makefile
```diff
@@ -1,14 +1,26 @@
+ifndef DEADFISH_DOCKER_NOSUDO
+dockercmd=sudo docker
+else
+dockercmd=docker
+endif
```
mparafin, line 5: This doesn't really change much, besides adding a bit of universality. It's ok, I understand the concept (and hey, it's a contribution, thanks :D), but I'm wondering whether things that anyway require additional local changes to work should be in the repo. Maybe a script for making docker work without sudo would make it more full. Idk, calling in @capturcus, what do you think?

piodul: The main benefit of this patch is that my local changes are restricted just to `~/.bashrc`. I only need to define an environment variable once, and suddenly I don't need to use sudo anymore. Otherwise I have to modify the `Makefile` and watch out in order not to include changes to it when doing `git commit`.

As for fixing the need to use sudo for docker... I don't think we can create a script that will work on all environments. The fix probably depends on the configuration of your OS.

Makefile
piodul: I'll add support for sudo-less usage of docker in `dfctl`, too.

As for `dfctl` vs. `make`, that's probably up to @capturcus to decide. Personally, I prefer `make` because I don't have to install anything (with sudo, of course!), and I stil get autocompletion.

capturcus: @piodul 's reasoning about `make` makes sens to me. Because of using a different shell I also don't get autocompletion with `dfctl`. We should probably migrate everything to `make`, but that's a task for a different PR. But this means that I will still merge this even without `dfctl` support.

Makefile
capturcus: Running docker with or without sudo depends on your system configuration and I want to enable both choices. This change does exactly that.

capturcus (approved):
lgtm

piodul:
Update: pushed support for the `DEADFISH_DOCKER_NOSUDO` env variable in the `dfctl` command.

mparafin:
image

PR #35, opened 2020-11-21
Agones 

capturcus:
*before compiling this you must rebuild the docker image*
- included agones in the development docker image
- included agones in our build system
- added scripts for automating agones related tasks
- added a simple server module for communicating with agones
- added agones files for launching a single server or a 2 server fleet


server/agones.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace agones {
+
+void Start();
+void Shutdown();
+void SetPlayers(int players);
+void SetPlaying();
+
+}; // agones
```
piodul: Nit: missing endline

server/agones.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <agones/sdk.h>
+#include <grpc++/grpc++.h>
+#include <iostream>
+#include <thread>
+
+static std::shared_ptr<agones::SDK> sdk;
+
+// send health check pings
+void DoHealth()
```
piodul: Make it `static`

server/agones.cpp
```diff
@@ -0,0 +1,131 @@
...
+		bool ok = sdk->Health();
+		if (!ok) {
+			std::cout << "health ping failed\n";
+		}
+		std::this_thread::sleep_for(std::chrono::seconds(2));
+	}
+}
+
+// watch GameServer Updates
+void WatchUpdates()
```
piodul: Make it `static`

server/agones.cpp
```diff
@@ -0,0 +1,131 @@
...
+	{
+		std::cerr << "Could not run GameServer(): " << status.error_message()
+				  << ". Exiting!\n";
+		exit(1);
+	}
+
+	std::cout << "GameServer name: " << gameserver.object_meta().name() << "\n"
+			  << std::flush;
+
+	health.join();
```
piodul: Doesn't this hang this thread? The `health` thread runs an infinite loop.

piodul: We should notify the `health` thread that we are stopping and that it should no longer be running. Here is a suggestion how to do it: https://stackoverflow.com/a/59756732/12851162

server/main.cpp
```diff
@@ -187,6 +192,9 @@ int main(int argc, const char* const argv[])
 
 	std::cout << "server started\n";
 
+	if (gameState.options["agones"].as<bool>())
+		new std::thread(agones::Start);
```
piodul: 🙄

server/agones.cpp
```diff
@@ -0,0 +1,131 @@
...
+			  << std::flush;
+	sdk->WatchGameServer([](const agones::dev::sdk::GameServer &gameserver) {
+		std::cout << "GameServer Update:\n"								   //
+				  << "\tname: " << gameserver.object_meta().name() << "\n" //
+				  << "\tstate: " << gameserver.status().state() << "\n"
+				  << std::flush;
+	});
+}
+
+void dfAgones::Start()
```
piodul: I don't think this function should be blocking. Its name suggests that it should return after Agones SDK's initialization finishes.

I'd move everything after line:

```c++
	std::cout << "GameServer name: " << gameserver.object_meta().name() << "\n"
			  << std::flush;
```

to the beginning of the `dfAgones::Shutdown()` function.

server/agones.cpp
```diff
@@ -0,0 +1,131 @@
...
+			  << "Getting the instance of the SDK.\n"
+			  << std::flush;
+	sdk = std::make_shared<agones::SDK>();
+
+	std::cout << "Attempting to connect...\n"
+			  << std::flush;
+	if (!sdk->Connect())
+	{
+		std::cerr << "Exiting!\n";
+		exit(1);
```
piodul: IMO, the function `dfAgones::Start()` should return a boolean value indicating if we successfully started the server.

piodul (changes requested):
(reviewed c++ code for now, didn't look at the YAMLs)

capturcus:
idea: agones::Start initializes the agones sdk and if it all succeeds launches the separate agones threads containing health checks and server state watches.

server/agones.cpp
capturcus: resolved this according to the comment below

server/main.cpp
capturcus: resolved this according to the comment below

server/agones.cpp
capturcus: resolved this according to the comment below

server/agones.cpp
capturcus: resolved this according to the comment below

server/agones.cpp
```diff
@@ -0,0 +1,111 @@
...
+	}
+}
+
+// watch GameServer Updates
+static void WatchUpdates()
+{
+	std::cout << "Starting to watch GameServer updates...\n"
+			  << std::flush;
+	sdk->WatchGameServer([](const agones::dev::sdk::GameServer &gameserver) {
+		std::cout << "GameServer Update:\n"								   //
```
piodul: Nit: There is an empty comment in this line, prepended by lots of whitespace characters

piodul (approved):

PR #37, opened 2020-12-19
FOV implementation

piodul:
Non-visible areas are now covered with shadows. Shadows are implemented for both polygonal and round collision masks.

Zrzut ekranu z 2020-12-20 21-00-40

### DIFF
--- a/client/dfclient/include/fov.hpp
+++ b/client/dfclient/include/fov.hpp
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <vector>
+#include <ncine/Vector2.h>
+
+namespace fov {
+
+/// A sequence of points which form a light-blocking chain.
+using chain = std::vector<ncine::Vector2f>;
+using segment = std::pair<ncine::Vector2f, ncine::Vector2f>;
+
+struct circle {
+	ncine::Vector2f pos;
+	float radius;
+};
+
+/// A mesh describing light-blocking parts of the level.
+class mesh {
+private:
+	std::vector<segment> _segments;
+	std::vector<circle> _circles;
+
+public:
+	void addChain(const chain& c);
+	void addChain(const circle& c);
+	std::vector<segment> calculateOutline(ncine::Vector2f origin) const;
+};
+
+}
--- a/client/dfclient/src/fov.cpp
+++ b/client/dfclient/src/fov.cpp
@@ -1,9 +1,21 @@
+#include <cassert>
+#include <cmath>
+#include <set>
+#include <queue>
+#include <functional>
+
 #include "fov.hpp"
 
 namespace fov {
 
+static std::vector<segment> deoverlap(ncine::Vector2f origin, std::vector<segment> segments);
+
 inline static double cross(ncine::Vector2f a, ncine::Vector2f b) {
 	return a.x * b.y - a.y * b.x;
+}
+
+inline static double dot(ncine::Vector2f a, ncine::Vector2f b) {
+	return a.x * b.x + a.y * b.y;
 }
 
 static bool isOrientedClockwise(const chain& c) {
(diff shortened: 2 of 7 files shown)


capturcus:
i just wanted you to know that you are a good person @piodul

piodul:
This is needed for the shadows to work: https://github.com/capturcus/deadfish-data/pull/1

piodul:
Self-review:

- Made the naming convention consistent with the one used in the code (previously it was mixed camel- and snake-case, now it's only camel case),
- Added comments explaining the details of the outline calculation and deoverlapping algorithm.

client/dfclient/include/fov.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <vector>
+#include <ncine/Vector2.h>
+
+namespace fov {
+
+/// A sequence of points which for a light-blocking chain.
```
capturcus:
```suggestion
/// A sequence of points which form a light-blocking chain.
```

client/dfclient/src/fov.cpp
```diff
@@ -0,0 +1,383 @@
...
+    return area >= 0.f;
+}
+
+// Adapted from Tim Voght's implementation of Paul Bourke's circle intersection
+// algorithm:
+// http://paulbourke.net/geometry/circlesphere/tvoght.c
+// Returns the intersection points of the circles. Always returns at most
+// two points.
+// `out` must point to an at-least-two-element array.
+static size_t circleIntersection(const circle& a, const circle& b, ncine::Vector2f* out) {
```
capturcus, line 45: This function does not account for the case `dist == a.radius + b.radius`. If there was such a case in a map the deoverlap algorithm would probably glitch a lot. Even though I don't think there is one I suggest that we support this

client/dfclient/src/fov.cpp
```diff
@@ -0,0 +1,383 @@
+#include <cassert>
```
capturcus, line 1: please indent with tabs

client/dfclient/src/fov.cpp
```diff
@@ -0,0 +1,383 @@
...
+#include <cmath>
+#include <set>
+#include <queue>
+#include <functional>
+
+#include "fov.hpp"
+
+namespace fov {
+
+static const bool doDeoverlap = true;
```
capturcus: I don't really see a case when we would need a not deoverlapped shape, I think we can remove this

client/dfclient/include/fov.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
```
capturcus, line 1: please indent with tabs

capturcus (changes requested):

client/dfclient/src/fov.cpp
piodul: This case shouldn't happen. Look at the place where this function is used, and at the circles that are intersected with each other.

We are given a circle with its center at point `C` and with radius `r`, and the player position `O`. Then, we construct another circle with center `C'` being the center of the segment `|CO|`, and radius `r' = |CO| / 2`. Unless `O` is inside the given circle (which shouldn't happen now because the player always collides with the circle given), those two circles will definitely overlap.

An illustration:

illustration

Even if the case you described happens, the function should just return two points which are the same. The outline will contain a degenerate segment (segment with coinciding ends). I doubt it will cause glitches, because the shadow cast by it shouldn't be visible at all (I may be wrong though, didn't test this).

client/dfclient/include/fov.hpp
piodul: Done

client/dfclient/src/fov.cpp
piodul: Removed

client/dfclient/src/fov.cpp
piodul: Done, changed indentation to tabs

client/dfclient/include/fov.hpp
piodul: Wait, I actually forgot to fix the indentation in the headers

piodul: Now it's done.

client/dfclient/src/fov.cpp
capturcus: Ok. The image convinces me

capturcus (approved):

PR #38, opened 2020-12-20
Client matchmaking

capturcus:
- choose between direct connection mode and matchmaking mode (and set default values) with an .ini file
- use a simple synchronous boost http server to connect to the matchmaker
- handle data received from the matchmaker
- minor fixes, including ui

### DIFF
--- a/client/dfclient/deadfish.ini.example
+++ b/client/dfclient/deadfish.ini.example
@@ -1,0 +1,6 @@
+[default]
+client_mode=matchmaker
+matchmaker_address=localhost
+matchmaker_port=8000
+direct_address=localhost
+direct_port=63987
--- a/client/dfclient/include/menu_state.hpp
+++ b/client/dfclient/include/menu_state.hpp
@@ -18,6 +18,7 @@ public:
 private:
 	bool TryConnect();
 	void ProcessMatchmakerData(std::string data);
+	void ShowMessage(std::string message);
 
 	Resources& _resources;
 };
--- a/client/dfclient/include/menu_state.hpp
+++ b/client/dfclient/include/menu_state.hpp
@@ -6,6 +6,10 @@
 
 class Resources;
 
+enum class ClientMode {
+	Matchmaker,
+	DirectConnect
+};
 struct MenuState
 	: public GameState
 {
@@ -19,6 +23,14 @@ private:
 	bool TryConnect();
 	void ProcessMatchmakerData(std::string data);
 	void ShowMessage(std::string message);
+	void MatchmakerLayout(char* buf);
+	void DirectConnectLayout(char* buf0, char* buf1);
+
+	ClientMode clientMode = ClientMode::Matchmaker;
+	std::string matchmakerAddress = "localhost";
+	std::string matchmakerPort = "8000";
+	std::string directAddress = "localhost";
+	std::string directPort = "63987";
 
 	Resources& _resources;
 };
--- a/client/dfclient/include/menu_state.hpp
+++ b/client/dfclient/include/menu_state.hpp
@@ -32,8 +32,8 @@ private:
 	std::string directAddress = "localhost";
 	std::string directPort = "63987";
 
-	char buf0[64];
-	char buf1[64];
+	char buf0[64] = {0};
+	char buf1[64] = {0};
 
 	Resources& _resources;
 };
--- a/client/dfclient/include/menu_state.hpp
+++ b/client/dfclient/include/menu_state.hpp
@@ -20,17 +20,20 @@ public:
 	StateType Update(Messages) override;
 
 private:
-	bool TryConnect();
+	void TryConnect();
 	void ProcessMatchmakerData(std::string data);
 	void ShowMessage(std::string message);
-	void MatchmakerLayout(char* buf);
-	void DirectConnectLayout(char* buf0, char* buf1);
+	void MatchmakerLayout();
+	void DirectConnectLayout();
 
 	ClientMode clientMode = ClientMode::Matchmaker;
 	std::string matchmakerAddress = "localhost";
 	std::string matchmakerPort = "8000";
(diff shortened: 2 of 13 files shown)


client/dfclient/src/http_client.cpp
```diff
@@ -0,0 +1,82 @@
...
+	return res.result_int();
+}
+
+int http::MatchmakerGet(std::string address, std::string port, std::string& ret)
+{
+	int retCode = 0;
+	try {
+		retCode = throwingMatchmakerGet(address, port, ret);
+	} catch (...) {
+		return -1;
```
piodul, line 80: How about printing the exception in the logs?

client/dfclient/src/http_client.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+	// Set up an HTTP GET request message
+	bhttp::request<bhttp::string_body> req{bhttp::verb::get, "/matchmake", 10};
+	req.set(bhttp::field::host, address);
+	req.set(bhttp::field::user_agent, BOOST_BEAST_VERSION_STRING);
+
+	// Send the HTTP request to the remote host
+	bhttp::write(stream, req);
+
+	// This buffer is used for reading and must be persisted
```
piodul: This comment is a little imprecise. Persisted, but for how long? What does "persisted" mean here?

client/dfclient/src/menu_state.cpp
```diff
@@ -14,40 +18,119 @@
...
+	}
+}
+
+void MenuState::ProcessMatchmakerData(std::string data) {
+	boost::property_tree::ptree tree;
+	std::stringstream ss;
+	ss << data;
+	boost::property_tree::read_json(ss, tree);
+
+	auto status = tree.get("status", "");
```
piodul, line 104: Is the server-matchmaker protocol documented somewhere?

client/dfclient/src/websocket.cpp
```diff
@@ -163,16 +163,20 @@ int WebSocketBeast::Connect(std::string& fullAddress) {
...
+			[](websocket::request_type& req)
+			{
+				req.set(http::field::user_agent,
+					std::string(BOOST_BEAST_VERSION_STRING) +
+						" websocket-client-coro");
+			}));
+		ws.handshake(host, "/");
+	} catch (std::exception e) {
+		return -1;
+	}
```
piodul, line 179: What's the reason for this change?

client/dfclient/src/http_client.cpp
capturcus: Actually read the comments copied from an example. Removed the "persisted" part. Everything else checks out

client/dfclient/deadfish.ini.example
```diff
@@ -0,0 +1,6 @@
+[default]
+client_mode=matchmaker
+matchmaker_address=localhost
+matchmaker_port=8000
+direct_address=localhost
+direct_port=63987
```
mparafin, line 6: The viable options are not documented anywhere. Maybe some comments here could be a good idea, as it is an example file? But anywhere would be ok, even in the [README.md](https://github.com/capturcus/deadfish/blob/master/README.md).

client/dfclient/include/menu_state.hpp
```diff
@@ -16,7 +20,20 @@ struct MenuState
...
+	void DirectConnectLayout();
+
+	ClientMode clientMode = ClientMode::Matchmaker;
+	std::string matchmakerAddress = "localhost";
+	std::string matchmakerPort = "8000";
+	std::string directAddress = "localhost";
+	std::string directPort = "63987";
+
+	char buf0[64];
+	char buf1[64];
```
mparafin: Is there no way to use less non-descriptive names? :D They seem to be indentical in functionality and one of them  is always a nickname, while second is __sometimes__ a server adress. Maybe they could be assigned a single role and named as such.

client/dfclient/src/menu_state.cpp
```diff
@@ -14,40 +18,119 @@
...
+{
+	nc::SceneNode &rootNode = nc::theApplication().rootNode();
+	auto res = nc::theApplication().appConfiguration().resolution;
+	auto text = new ncine::TextNode(&rootNode, _resources.fonts["comic"].get());
+	text->setString(message.c_str());
+	text->setPosition(res.x * 0.5f, res.y * 0.30f);
+	text->setScale(0.5f);
+	text->setColor(0, 0, 0, 255);
+	_resources._tweens.push_back(CreateTextTween(text));
+}
```
mparafin: I still don't like the idea of leaking TextNodes just to show them xd I know it's copied, just nitpicking. Maybe I'll refactor it someday. :D

mparafin (approved):
L(V)GTM :) You have my approve, all the comments are just nits of the kind of code hygiene. For me the code is mergeable even as it is :)

client/dfclient/src/menu_state.cpp
capturcus: It is now in the README

client/dfclient/src/websocket.cpp
capturcus: Adding a try/catch block serves to return a value on error instead of crashing with an uncaught exception

client/dfclient/deadfish.ini.example
capturcus: Added coments in the ini example file

client/dfclient/include/menu_state.hpp
capturcus: Renamed buffers and assigned roles to them

client/dfclient/src/menu_state.cpp
capturcus: True, I don't know why did I commit this, sorry. Added deleting the text pointer when the tween reaches 0

PR #39, opened 2021-01-17
Movables refactor

capturcus:
Introduced a concept of a "movable" thing, that is something that:
- can move on the server and the movement is reflected on the client
- supports server side line of sight (although I ended up doing this more or less manually)
- can be tracked with an ID so that lerping and client side sprite management can be automatically performed

Added a movable component to the flatbuffer definitions, implemented handling of the new structure server and client side.
There are no visible changes because of this apart from ink particles and mob manipulators tweening in and out of existence with alpha. However this should make adding new movable things a little easier.


common/geometry.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <iostream>
+
```
mparafin: ?? Probably a debugging artifact. If we don't need it here, please remove :)

server/deadfish.hpp
```diff
@@ -65,8 +81,20 @@ struct Collideable {
...
+struct Movable {
+	FlatBuffGenerated::Vec2 pos;
+	uint16_t movableID;
+	float angle;
+
+	virtual std::unique_ptr<FlatBuffGenerated::MovableComponent> fbMovable();
+};
+
+struct CollideableMovable
+	: public Collideable, public Movable {
```
mparafin, line 93: It's starting to look a bit spooky :D But as long as we don't have more object properties as these two (and I don't suppose we're going to) it's ok.

mparafin (approved):
Very nice, very nice :)

client/dfclient/include/gameplay_state.hpp
```diff
@@ -80,20 +86,28 @@ struct GameplayState
 
 	void updateRemainingText(uint64_t remainingFrames);
 	void updateShadows();
+	void updateHovers(ncine::Vector2f mouseCoords, float radiusSquared);
+
+	template<typename T, typename F>
+	void processMovable(std::map<uint16_t, T>& map, const FlatBuffGenerated::MovableComponent* comp,
+		F createMovableFunc);
+	template<typename T>
+	void deleteUnusedMovables(std::map<uint16_t, T>& map);
```
piodul, line 101: You put declarations of those function templates into the header, but put their definitions into a single cpp file (`gameplay_state.cpp`). Currently, those functions aren't going to work outside the `gameplay_state.cpp` file.

Please put definitions of those templates into the `gameplay_state.hpp` file, for example after the `GameplayState` struct definition.

client/dfclient/src/gameplay_state.cpp
```diff
@@ -416,48 +449,35 @@ void GameplayState::ProcessWorldState(const void* ev) {
...
-	}
+		auto manipulatorData = worldState->mobManipulators()->Get(i);
+		processMovable(this->manipulators, manipulatorData->movable(), [&](){
+			auto manipTexture = manipulatorData->type() == FlatBuffGenerated::MobManipulatorType_Dispersor ?
+				"dispersor.png" : "attractor.png";
+			auto sprite = std::make_unique<ncine::Sprite>(this->cameraNode.get(), _resources.textures[manipTexture].get());
+			sprite->setLayer((unsigned short) Layers::MOB_MANIPULATORS);
+			Manipulator ret;
+			ret.sprite = std::move(sprite);
+			return std::move(ret);
```
piodul: `std::move` is not needed when returning, and actually worsens performance. Please just do `return ret;`

This is called a [pessimizing move](https://stackoverflow.com/questions/62061433/how-to-avoid-the-pessimizing-move-warning-of-nrvo).

client/dfclient/src/gameplay_state.cpp
```diff
@@ -305,47 +305,94 @@ void GameplayState::ProcessSimpleServerEvent(const void* ev) {
...
+		// we see it for the first time
+		T newMov = createMovableFunc();
+		newMov.lerp.bind(static_cast<ncine::DrawableNode*>(newMov.sprite.get()));
+
+		//fade in
+		newMov.sprite->setAlpha(1);
+		newMov.tween = CreateAlphaTransitionTween(newMov.sprite.get(), 1, 255, MOB_FADEIN_TIME);
+		newMov.movableID = comp->ID();
+		map[comp->ID()] = std::move(newMov);
+		it = map.find(comp->ID());
```
piodul: Nit: Use `insert` to avoid double lookup:

```c++
it = map.insert(std::move(newMov)).first;
```

client/dfclient/src/gameplay_state.cpp
```diff
@@ -416,48 +449,35 @@ void GameplayState::ProcessWorldState(const void* ev) {
...
 			newInk.sprite->setLayer((unsigned short) Layers::INK_PARTICLES);
 			newInk.sprite->setScale(120./330.); // todo: fix my life
-
-			inkParticles.insert({ink->inkID(), std::move(newInk)});
-			inkItr = inkParticles.find(ink->inkID());
-			inkItr->second.lerp.bind(inkItr->second.sprite.get());
-		}
-		inkItr->second.lerp.setupLerp(ink->pos()->x(), ink->pos()->y(), inkItr->second.sprite->rotation());
-		inkItr->second.seen = true;
+			return std::move(newInk);
```
piodul: Don't `std::move` when returning

piodul (approved):

server/deadfish.hpp
capturcus: This particular combo is important. This struct overloads the function fbMovable to handle copying data from the physics engine to the movable structure

client/dfclient/include/gameplay_state.hpp
capturcus: Moved the functions and their dependencies into the header file

client/dfclient/src/gameplay_state.cpp
capturcus: Removed `std::move`s. Thanks for the read

PR #40, opened 2021-02-22
Destination markers

mparafin:
A tiny and basic destination marker implementation. Doesn't disappear when reached, but it's partially intentional - the character will still try to reach the destination when displaced by other mobs.

### DIFF
--- a/client/dfclient/src/gameplay_state.cpp
+++ b/client/dfclient/src/gameplay_state.cpp
@@ -565,14 +565,23 @@ void GameplayState::OnMouseButtonPressed(const ncine::MouseEvent &event) {
 		// move
 		const float screenWidth = ncine::theApplication().width();
 		const float screenHeight = ncine::theApplication().height();
-		const float serverX = (this->mySprite->position().x + (event.x - screenWidth / 2) * 1.5f) * PIXELS2METERS;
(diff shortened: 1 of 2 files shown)


client/dfclient/src/gameplay_state.cpp
```diff
@@ -565,14 +565,23 @@ void GameplayState::OnMouseButtonPressed(const ncine::MouseEvent &event) {
...
 		flatbuffers::FlatBufferBuilder builder;
 		auto pos = FlatBuffGenerated::Vec2(serverX, serverY);
 		auto cmdMove = FlatBuffGenerated::CreateCommandMove(builder, &pos);
 		auto message = FlatBuffGenerated::CreateClientMessage(builder, FlatBuffGenerated::ClientMessageUnion_CommandMove, cmdMove.Union());
 		builder.Finish(message);
 		SendData(builder);
+
+		auto player = dynamic_cast<ncine::AnimatedSprite*>(this->mySprite);
+		// if player is killing sb at the moment, don't create destination marker
+		if (player->animationIndex() == 2) return;
```
capturcus:
```suggestion
		if (player->animationIndex() == FlatBuffGenerated::MobState_Attack) return;
```

capturcus (changes requested):

capturcus (approved):

PR #41, opened 2021-04-09
Loading screen

mparafin:
Fancy loading screen in place of looking at uninitialized graphical buffer (which is fun, but a bit unprofessional). Hoo!

image

### DIFF
--- a/client/dfclient/src/loading_state.cpp
+++ b/client/dfclient/src/loading_state.cpp
@@ -58,26 +58,26 @@ LoadingState::LoadingState(Resources& r) : _resources(r) {
 	_resources.textures["blackpixel.png"] = std::make_unique<ncine::Texture>((rootPath + "loading/blackpixel.png").c_str());
 
 	// init loading text
-	_loading_text = std::make_unique<ncine::TextNode>(&rootNode, _resources.fonts["comic"].get());
-	_loading_text->setColor(ncine::Color::White);
-	_loading_text->setPosition(res.x*0.5f, res.y*0.5f);
-	_loading_text->setString("Loading... 0%");
+	_loadingText = std::make_unique<ncine::TextNode>(&rootNode, _resources.fonts["comic"].get());
+	_loadingText->setColor(ncine::Color::White);
+	_loadingText->setPosition(res.x*0.5f, res.y*0.5f);
+	_loadingText->setString("Loading... 0%");
 
 	// init loading graphics
-	_fillup_sprite = std::make_unique<ncine::Sprite>(&rootNode, _resources.textures["fillup"].get());
-	_fillup_sprite->setPosition(res.x*0.5f, res.y*0.6f);
-	_fillup_sprite->setLayer(1);
+	_fillupSprite = std::make_unique<ncine::Sprite>(&rootNode, _resources.textures["fillup"].get());
+	_fillupSprite->setPosition(res.x*0.5f, res.y*0.6f);
+	_fillupSprite->setLayer(1);
 
 	_curtain = std::make_unique<ncine::MeshSprite>(&rootNode, _resources.textures["blackpixel.png"].get());
 	std::vector<ncine::Vector2f> vertices(4);	
-	vertices[0].x = _fillup_sprite->position().x - _fillup_sprite->width()/2;
-	vertices[0].y = _fillup_sprite->position().y - _fillup_sprite->height()/2;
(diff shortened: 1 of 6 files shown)


client/dfclient/src/loading_state.cpp
```diff
@@ -0,0 +1,164 @@
...
+	while (file)
+	{
+		auto absPath = rootPath + std::string(TEXTURES_PATH) + "/" + std::string(file);
+		if (ncine::FileSystem::isFile(absPath.c_str())) {
+			_total += fileSize(absPath);
+			_textureFilenames.push_back(absPath);
+		}
+		file = textureDir.readNext();
+	}
+	textureDir.close();
```
capturcus: maybe abstract this to a function?

client/dfclient/src/loading_state.cpp
```diff
@@ -0,0 +1,164 @@
...
+	file = soundDir.readNext();
+	while (file) {
+		auto absPath = rootPath.data() + std::string(SOUNDS_PATH) + "/" + std::string(file);
+		if (ncine::FileSystem::isFile(absPath.c_str())) {
+			_total += fileSize(absPath);
+			_soundFilenames.push_back(absPath);
+		}
+		file = soundDir.readNext();
+	}
+
```
capturcus, line 46: the textureDir is closed, but soundDir never is. does that make any difference? please decide on one of the variants

client/dfclient/src/loading_state.cpp
```diff
@@ -0,0 +1,164 @@
...
+	unsigned char alpha = 55 + 2*_percent_loaded;
+	_loadingText->setAlpha(alpha);
+
+	// update the loading graphics
+	_curtain->setPosition(0, 1.2*_percent_loaded);
+
+	// manage the ending
+	if (_percent_loaded == 100) {
+		if (_fadeoutTween == nullptr) {
+			// stretch curtain to cover the whole screen
```
capturcus, line 120: to be honest I'm not sure whether I'm seeing the curtain effect at all. but overall the game looks much better and doesn't crash so that's ok :D

client/dfclient/src/loading_state.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+LoadingState::~LoadingState() {
+	// init soundbuffers
+	_resources._killSoundBuffer = std::move(_resources._sounds["amongus-kill.wav"]);
+	_resources._sounds.erase("amongus-kill.wav");
+	_resources._killSound = std::make_unique<ncine::AudioBufferPlayer>(_resources._killSoundBuffer.get());
+
+	_resources._goldfishSoundBuffer = std::move(_resources._sounds["mario-powerup.wav"]);
+	_resources._sounds.erase("mario-powerup.wav");
+	_resources._goldfishSound = std::make_unique<ncine::AudioBufferPlayer>(_resources._goldfishSoundBuffer.get());
```
capturcus: why is there some initialization logic in the _destructor_

capturcus (changes requested):
I absolutely love the loading screen :smile:

capturcus:
Please restore the blackpixel.png file to its original location and use it from there. It is also needed in other places than the loading screen.

client/dfclient/src/loading_state.cpp
mparafin: The curtain then is used to do the fade-out effect, it's a black rectangle that is becoming less and less transparent. It doesn't go anywhere after that, just disappears :D

client/dfclient/src/loading_state.cpp
mparafin: Because this way we know it will be executed and that it will be executed lastly. It cannot be in the Update method (_maybe_ it could be placed after an if-gate that makes sure everything is finished loading, but I don't like the idea of checking if we are done after every file) and the destructor of the LoadingState seems like a good transition place.

client/dfclient/src/loading_state.cpp
mparafin: That's right, my bad. Added

client/dfclient/src/loading_state.cpp
capturcus: >I don't like the idea of checking if we are done after every file

we do it anyways
```
if (_percent_loaded == 100) {
```
I'm not going to make you commit this trivial change yourself, but I will change it. Initialization logic in the destructor seems like a very bad idea to me

capturcus (approved):

PR #42, opened 2021-04-09
Ink cloud visibility fix

mparafin:
Made ink clouds not obscure each other while still obscuring vision of mobs.

image


capturcus (approved):
nigga

PR #43, opened 2021-04-15
server and client websocket closing

capturcus:
this makes repeated connections to a game that is already in progress handle nicely
there still is a race somewhere but I've seen it only once so _it cannot be that bad_

### DIFF
--- a/client/dfclient/src/menu_state.cpp
+++ b/client/dfclient/src/menu_state.cpp
@@ -41,7 +41,6 @@ void MenuState::ShowMessage(std::string message)
 		[text] (tweeny::tween<int>& t, int v) -> bool {
 			auto textColor = text->color();
 			text->setColor(textColor.r(), textColor.g(), textColor.b(), v);
-			std::cout << "v " << v << "\n";
 			if (v == 0)
 				delete text;
 			return false;
@@ -61,6 +60,8 @@ MenuState::MenuState(Resources& r) : _resources(r) {
 		TextCreator textCreator(r);
 		std::cout << "menu game in progress\n";
 		ShowMessage("game already in progress");
+		webSocketManager._ws->Close();
+		webSocketManager._ws = nullptr;
 		gameData.gameInProgress = false;
 	}
 	boost::property_tree::ptree pt;
@@ -91,8 +92,8 @@ MenuState::MenuState(Resources& r) : _resources(r) {
 void MenuState::TryConnect() {
 	gameData.serverAddress = "ws://" + gameData.serverAddress;
 	std::cout << "server " << gameData.serverAddress << ", my nickname " << gameData.myNickname << "\n";
-	gameData.socket = CreateWebSocket();
-	int ret = gameData.socket->Connect(gameData.serverAddress);
+	CreateWebSocket();
+	int ret = webSocketManager._ws->Connect(gameData.serverAddress);
 	if (ret < 0) {
 		std::cout << "socket->Connect failed " << ret << "\n";
 		this->ShowMessage("connecting to server failed");
--- a/client/dfclient/src/menu_state.cpp
+++ b/client/dfclient/src/menu_state.cpp
@@ -60,8 +60,8 @@ MenuState::MenuState(Resources& r) : _resources(r) {
 		TextCreator textCreator(r);
 		std::cout << "menu game in progress\n";
 		ShowMessage("game already in progress");
-		webSocketManager._ws->Close();
-		webSocketManager._ws = nullptr;
+		GlobalWebsocket->Close();
+		GlobalWebsocket = nullptr;
 		gameData.gameInProgress = false;
 	}
 	boost::property_tree::ptree pt;
@@ -93,7 +93,7 @@ void MenuState::TryConnect() {
 	gameData.serverAddress = "ws://" + gameData.serverAddress;
 	std::cout << "server " << gameData.serverAddress << ", my nickname " << gameData.myNickname << "\n";
 	CreateWebSocket();
-	int ret = webSocketManager._ws->Connect(gameData.serverAddress);
+	int ret = GlobalWebsocket->Connect(gameData.serverAddress);
 	if (ret < 0) {
 		std::cout << "socket->Connect failed " << ret << "\n";
 		this->ShowMessage("connecting to server failed");
--- a/client/dfclient/src/util.cpp
+++ b/client/dfclient/src/util.cpp
@@ -16,7 +16,7 @@ void SendData(flatbuffers::FlatBufferBuilder& builder) {
 	auto size = builder.GetSize();
 	auto str = std::string(data, data + size);
 
-	gameData.socket->Send(str);
+	webSocketManager._ws->Send(str);
 }
 
 std::vector<nc::Vector2f> createArcTexels(float outerRadius, float innerRadius, int degrees)
--- a/client/dfclient/src/util.cpp
+++ b/client/dfclient/src/util.cpp
@@ -16,7 +16,7 @@ void SendData(flatbuffers::FlatBufferBuilder& builder) {
 	auto size = builder.GetSize();
 	auto str = std::string(data, data + size);
 
-	webSocketManager._ws->Send(str);
+	GlobalWebsocket->Send(str);
 }
 
 std::vector<nc::Vector2f> createArcTexels(float outerRadius, float innerRadius, int degrees)
--- a/client/dfclient/src/websocket.cpp
+++ b/client/dfclient/src/websocket.cpp
@@ -31,6 +31,7 @@ class WebSocketEmscripten
 public:
 	int Connect(std::string& address) override;
 	bool Send(std::string& data) override;
+	void Close() override;
 	virtual ~WebSocketEmscripten() {}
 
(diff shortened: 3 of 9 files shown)


server/websocket.cpp
```diff
@@ -141,26 +141,48 @@ class DfWebsocket : public std::enable_shared_from_this<DfWebsocket> {
...
+
+void dfws::Close(Handle hdl)
+{
+    std::cout << "closing handle " << hdl << "\n";
+    auto it = sockets.begin();
+    for (; it != sockets.end(); it++) {
+        if ((*it)->socketID_ == hdl) {
+            break;
+        }
+    }
```
piodul: Suggestion: change to:

```c++
auto it = std::find_if(sockets.begin(), sockets.end(), [&] (const auto& s) {
    return s.socketID_ == hdl;
});
```

client/dfclient/src/websocket.cpp
```diff
@@ -183,19 +199,37 @@ int WebSocketBeast::Connect(std::string& fullAddress) {
 	DoRead();
 
 	new std::thread([&](){
-		ioc.run();
+		try {
+			wsb->ctx.ioc.run();
+		} catch (const std::exception& e) {
+			if (e.what() == std::string("Operation canceled")) {
```
piodul: You should match on the correct exception type in the `catch` clause, not compare the description string.

client/dfclient/src/websocket.cpp
```diff
@@ -183,19 +199,37 @@ int WebSocketBeast::Connect(std::string& fullAddress) {
...
+			}
+			throw e;
+		}
+		std::cout << "ioc quitting\n";
 	});
 
 	return 0;
 }
 
 bool WebSocketBeast::Send(std::string& data) {
```
piodul: Why is the argument to this function a mutable reference? The string behind the reference is never modified.

client/dfclient/src/websocket.cpp
```diff
@@ -183,19 +199,37 @@ int WebSocketBeast::Connect(std::string& fullAddress) {
...
+	WebSocketBeast* wsb = (WebSocketBeast*) webSocketManager._ws.get();
+	wsb->ctx.ws.write(net::buffer(std::string(data)));
+}
+
+void WebSocketBeast::Close() {
+	WebSocketBeast* wsb = (WebSocketBeast*) webSocketManager._ws.get();
+	try {
+		wsb->ctx.ws.close(websocket::close_code::normal);
+	} catch (...) {
+		std::cout << "ws.close error \n";
```
piodul: It would be nice to print the exception. You could introduce two `catch` clauses:

```c++
try {
    wsb->ctx.ws.close(websocket::close_code::normal);
} catch (const std::exception& ex) {
    std::cout << "exception occured while closing websocket: " << ex.what() << std::endl;
} catch (...) {
    std::cout << "an unknown exception occured while closing websocket" << std::endl;
}
```

client/dfclient/src/util.cpp
```diff
@@ -16,7 +16,7 @@ void SendData(flatbuffers::FlatBufferBuilder& builder) {
 	auto size = builder.GetSize();
 	auto str = std::string(data, data + size);
 
-	gameData.socket->Send(str);
+	webSocketManager._ws->Send(str);
```
piodul: What it the purpose of the `webSocketManager`? Everybody seems to only access its `_ws` field and nothing else. Can we get rid of it and just use the websocket directly, or add some methods so that it does something useful?

server/websocket.cpp
```diff
@@ -141,26 +141,48 @@ class DfWebsocket : public std::enable_shared_from_this<DfWebsocket> {
...
-void dfws::SendData(Handle hdl, const std::string& data)
+std::shared_ptr<DfWebsocket> getSocketForHandle(dfws::Handle hdl)
 {
     for (auto& s : sockets) {
         if (s->socketID_ == hdl) {
-            s->Send(data);
-            return;
+            return s;
         }
     }
```
piodul: `std::find_if` could be used here, too

client/dfclient/src/websocket.cpp
mparafin: Especially so, that it throws every time when game is already in progress. I wonder if it's intended, although it does work :D

mparafin: Especially so, that it throws every time when game is already in progress. I wonder if it's intended, although it does work :D

client/dfclient/src/menu_state.cpp
```diff
@@ -91,8 +92,8 @@ MenuState::MenuState(Resources& r) : _resources(r) {
 void MenuState::TryConnect() {
 	gameData.serverAddress = "ws://" + gameData.serverAddress;
 	std::cout << "server " << gameData.serverAddress << ", my nickname " << gameData.myNickname << "\n";
-	gameData.socket = CreateWebSocket();
-	int ret = gameData.socket->Connect(gameData.serverAddress);
+	CreateWebSocket();
+	int ret = webSocketManager._ws->Connect(gameData.serverAddress);
 	if (ret < 0) {
```
mparafin: The race is here - adding `sleep(100ms)` after `Connect()` seemed to solve the problem.

client/dfclient/src/util.cpp
mparafin: Functionality: https://github.com/capturcus/deadfish/blob/a56b3db6c6e3560ff7e9bb9ad0534208d1207837/client/dfclient/src/websocket.cpp#L8
Use: https://github.com/capturcus/deadfish/blob/a56b3db6c6e3560ff7e9bb9ad0534208d1207837/client/dfclient/src/state_manager.cpp#L65

mparafin (changes requested):
Apart from the race, seems to be working quite ok. The code itself is a bit messy, but it works - I think that a refactor would be in order, but that's a different task pretty far in the backlog. After resolving the race (and considering piodul's remarks) lgtm.

client/dfclient/src/util.cpp
piodul: OK, `GetMessages` is useful, but it can just be a function which takes `_ws` as an additional argument. I don't see why this has to be a struct. It's ugly if "guts" of such manager are open for everybody to use.

client/dfclient/src/menu_state.cpp
piodul: Please use proper synchronization to fix it, though (e.g. mutexes, condition variables, semaphores).

mparafin: @piodul Don't get me wrong, it was just a note to pinpoint the problem, not a fix suggestion xD

client/dfclient/src/websocket.cpp
capturcus: It turned out that this wasn't needed at all, removed.

client/dfclient/src/websocket.cpp
capturcus: Changed to const. I see that good practices from Rust are making their way in here :P

palpatine-good

capturcus: It turns out that the emscripten websocket send function does not mark its argument as const and that's why it was so. I used a cast and assumed that javascript is not that retarded

client/dfclient/src/websocket.cpp
capturcus: Changed the exception handling to what @piodul suggested.

client/dfclient/src/util.cpp
capturcus: Got rid of the WebSocketManager. I think that originally there was some purpose to it but it was lost after a few refactors.

client/dfclient/src/menu_state.cpp
capturcus: According to my research the race should not be present, even though I did nothing in the code to mitigate it. After the call to `handshake` finished successfully the socket should be immediately available to send data. The only thing that might be racing here is if there is something happening within `ioc.run` that makes the websocket ready to send messages and there is no way to synchronize anything within an external function call except for a `sleep`.
After some testing I have **not** observed the race happening on my machine. It is possible that the racing behaviour was introduced in a different PR.

client/dfclient/src/websocket.cpp
```diff
@@ -194,42 +195,40 @@ int WebSocketBeast::Connect(std::string& fullAddress) {
...
+	WebSocketBeast* wsb = (WebSocketBeast*) GlobalWebsocket.get();
 	wsb->ctx.ws.write(net::buffer(std::string(data)));
 }
 
 void WebSocketBeast::Close() {
-	WebSocketBeast* wsb = (WebSocketBeast*) webSocketManager._ws.get();
+	WebSocketBeast* wsb = (WebSocketBeast*) GlobalWebsocket.get();
 	try {
-		wsb->ctx.ws.close(websocket::close_code::normal);
+    	wsb->ctx.ws.close(websocket::close_code::normal);
```
piodul: Nit: this line is indented with spaces, while the rest is indented with tabs.

piodul (approved):

client/dfclient/src/websocket.cpp
capturcus: nice catch

PR #44, opened 2021-04-16
Graceful closing of the GameplayState so that returning to menu on disconnect doesn't crash the client

capturcus:


capturcus:
EDIT: this is already done
The timer text needs to be deleted though. I should probably refactor this to a unique_ptr before this gets merged

client/dfclient/include/websocket.hpp
```diff
@@ -10,13 +10,14 @@
...
 	virtual int Connect(std::string& address) = 0;
-	virtual bool Send(std::string& data) = 0;
+	virtual void Send(std::string& data) = 0;
 	virtual void Close() = 0;
 	virtual ~WebSocket() {}
 
 	// TODO use lock-free queue instead?
 	std::mutex mq_mutex;
 	bool toBeOpened = false;
+	bool connectionClosed;
```
piodul: This doesn't seem to be initialized in the constructor. Please change it to the following:

```c++
bool connectionClosed = false;
```

client/dfclient/src/gameplay_state.cpp
```diff
@@ -432,6 +431,15 @@ GameplayState::GameplayState(Resources& r) : _resources(r), _deathReportProcesso
 }
 
 GameplayState::~GameplayState() {
+	// the order of these is important
```
piodul, line 434: Maybe you could elaborate here how those should be ordered and why the order is important?

mparafin (approved):
:+1:

capturcus:
I'm pretty sure this will have to be refactored after merging #43 . Please review but I will not merge until this is rebased onto `sockets` and tested. It is possible that some code here introduces a race.

piodul (approved):

PR #45, opened 2021-04-16
zigzags 0.1

capturcus:


server/raycasting.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return fovCb.data.closest && fovCb.data.closest->GetBody() == c.body;
+}
+
+bool mobSeePoint(Mob &m, const b2Vec2 &point, bool ignoreMobs, bool ignorePlayerwalls, bool ignoreHidingSpots)
+{
+	if (b2Distance(m.body->GetPosition(), point) == 0.0f)
+		return true;
+    struct Data {
+        float minfraction = 1.f;
+    };
```
piodul: You are mixing tabs and spaces...

common/geometry.hpp
```diff
@@ -1,5 +1,13 @@
 #pragma once
 
+static float normalizeAngle(float angle) {
```
piodul: This function is not used anywhere - it's dead code, so please remove it.

server/raycasting.hpp
```diff
@@ -0,0 +1,19 @@
...
+struct FOVCallback
+	: public b2RayCastCallback
+{
+	float32 ReportFixture(b2Fixture *fixture, const b2Vec2 &point, const b2Vec2 &normal, float32 fraction)
+	{
+        return f(fixture, point, normal, fraction, data);
+	}
+    std::function<float(b2Fixture*, const b2Vec2 &, const b2Vec2 &, float32, S& data)> f;
+    S data;
+};
```
piodul, line 21: I think `S data` and the template parameter is not needed here. Instead of defining a local structure `Data`, you could just allocate all variables on stack and capture them by reference in the lambda.

If you wanted to go further and avoid double virtual calls (`FOVCallback::ReportFixture` is one, and `std::function::operator()` must do a virtual call underneath), you could convert this structure to this:

```c++
template<typename F>
class FOVCallback final : public b2RayCastCallback
{
private:
    F _f;
public:
    FOVCallback(F&& f) : _f(std::move(f)) {}

    float32 ReportFixture(b2Fixture *fixture, const b2Vec2 &point, const b2Vec2 &normal, float32 fraction) override {
        return _f(fixture, point, normal, fraction);
    }
};
```

Then, it should be possible to use it like this:

```c++
float minFraction = 1.f;
auto fovCb = FOVCallback([&] (/* arguments */) {
    if (something > 42) {
        minFraction = (float)something;
    }
});

gameState.b2world->RayCast(&fovCb, m.body->GetPosition(), point);

File: server/raycasting.cpp [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+		if (ignoreMobs && dynamic_cast<Mob*>(collideable))
+			return 1.f;
+        if (ignorePlayerwalls && dynamic_cast<PlayerWall*>(collideable))
+			return 1.f;
+		if (ignoreHidingSpots && dynamic_cast<HidingSpot*>(collideable))
+			return 1.f;
+		if (fraction < data.minfraction)
+		{
+			data.minfraction = fraction;
+		}
```
piodul: More explicitly:

```c++
data.minFraction = std::min(data.minFraction, fraction);
```

server/mobs.cpp
```diff
@@ -176,9 +180,11 @@ void Civilian::collisionResolution() {
 		auto spawnPos = g2b(navpoint->position);
 
 		// not where we were currently going but somewhere we can go immediately from here
-		if (this->currentNavpoint != spawnName && mobSeePoint(*this, spawnPos, true)) {
+		if (!navpoint->isplayerspawn && spawnName != this->currentNavpoint && mobSeePoint(*this, spawnPos, true)) {
```
piodul: Why is `spawnName` named like that? It's a name of a navpoint, which is not necessarily a spawn point. It's confusing.

server/raycasting.cpp
capturcus: 😢😭🤦🤮

server/raycasting.hpp
capturcus: There really are no performance requirements to optimize virtual calls, but I guess it's more elegant that way. Refactored. FOVCallback didn't even have to be a template.

server/mobs.cpp
capturcus: refactored to `navpointName`

server/raycasting.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+bool mobSeePoint(Mob &m, const b2Vec2 &point, bool ignoreMobs, bool ignorePlayerwalls, bool ignoreHidingSpots)
+{
+	if (b2Distance(m.body->GetPosition(), point) == 0.0f)
+		return true;
+	float min_fraction = 1.f;
+	auto fovCb = FOVCallback([ignoreMobs, ignorePlayerwalls, ignoreHidingSpots, &min_fraction]
+		(b2Fixture *fixture, UNUSED const b2Vec2 &point, UNUSED const b2Vec2 &normal,
+	float32 fraction){
+	auto collideable = (Collideable *)fixture->GetBody()->GetUserData();
```
piodul, line 34: These lines should be indented more.

server/raycasting.cpp
```diff
@@ -0,0 +1,47 @@
...
+		// on return 1.f the currently reported fixture will be ignored and the raycast will continue
+		if (ignoreMobs && dynamic_cast<Mob*>(collideable))
+			return 1.f;
+	if (ignorePlayerwalls && dynamic_cast<PlayerWall*>(collideable))
```
piodul, line 38: This line should be indented more.

server/raycasting.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+class FOVCallback final : public b2RayCastCallback
+{
+private:
+	RaycastingCallback _f;
+public:
+	FOVCallback(RaycastingCallback&& f) : _f(std::move(f)) {}
+
+	float32 ReportFixture(b2Fixture *fixture, const b2Vec2 &point, const b2Vec2 &normal, float32 fraction) override {
+	return _f(fixture, point, normal, fraction);
```
piodul, line 19: This line should be indented more.

PR #46, opened 2021-04-17
Change the civilian pathfinding algorithm so that we don't have to manually create path graphs in the map editor

capturcus:


capturcus:
ymm maybe let's merge zigzags first

